PR #25, opened 2025-04-14
preflight

wenxie-amd:


tools/preflight/preflight_perf_test.py
```diff
@@ -0,0 +1,945 @@
...
+                        log(f"{hostname:<{max_len}} {node_id:<5} {rank:<5} {' '.join(formatted_values)}")
+                        f.write(f"| {hostname} | {node_id} | {rank} | {' | '.join(formatted_values)}|\n")
+                    f.write(f"\n")
+
+                    if not args.plot:
+                        continue
+
+                log(f"=======Plot IntraNode {case_name} Bandwidth=======")
+                with open(args.markdown_file, "a", encoding="utf-8") as f:
+                    f.write(f"=======Plot InterNode {case_name} Bandwidth=======\n")
```
lhzhang333: should be "... Plot IntraNode ..."

wenxie-amd: Fixed

PR #28, opened 2025-04-18
feat(megatron): Align TFLOPs calculation for megatron

Xiaoming-AMD:
Align TFLOPs calculation

### DIFF
--- a/primus/modules/trainer/megatron/trainer.py
+++ b/primus/modules/trainer/megatron/trainer.py
@@ -160,59 +160,214 @@ from primus.modules.trainer.base_trainer import BaseTrainer
 
 from .utils import set_wandb_writer_patch
 
-# The earliest we can measure the start time.
-_TRAIN_START_TIME = time.time()
-
 
 def num_floating_point_operations(args, batch_size):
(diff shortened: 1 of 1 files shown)


primus/modules/trainer/megatron/trainer.py
```diff
@@ -165,6 +164,57 @@
...
+            # MLP.
+            + ((args.ffn_hidden_size / args.hidden_size) * num_experts_routed_to * gated_linear_multiplier)
+            # Shared Experts.
+            + ((shared_expert_ffn_hidden_size / args.hidden_size) * gated_linear_multiplier)
+            # Logit.
+            + (args.padded_vocab_size / (2 * args.num_layers * args.hidden_size))
+        )
+    )
+
+
```
wenxie-amd: 是不是基于megatron最新的那个版本，然后把下面的 /2 去掉就行了？新的版本支持了moe_layer_freq
                    # Only half of the attention matrix is non-zero and needs to be multiplied with V.
                    + (args.seq_length / args.hidden_size / 2)

PR #79, opened 2025-06-18
feat(tp-overlap): add te backend and support tp overlap for megatron.

zhenhuang12:


primus/modules/trainer/megatron/trainer.py
```diff
@@ -1416,7 +1441,7 @@ def get_e2e_base_metrics():
                     active=args.profile_step_end - args.profile_step_start,
                     repeat=1,
                 ),
-                on_trace_ready=torch.profiler.tensorboard_trace_handler(args.tensorboard_dir),
+                on_trace_ready=torch.profiler.tensorboard_trace_handler("/workspace/Primus/profile"),
```
lhzhang333: Better to keep it configurable?

zhenhuang12: my bad! fixed!

PR #95, opened 2025-06-25
Add Tracer to primus megatron trainer

haishuok0525:


primus/modules/trainer/megatron/trainer.py
```diff
@@ -1225,7 +1228,7 @@ def run(self, *args, **kwargs):
         non_loss_data_func = None
         if not args.skip_train:
             log_rank_0("training ...")
-
+            self.tracer.start_training(self.model)
```
wenxie-amd, line 1231: Will this tracer impact performance? Tracing tools usually come with a argument — should we add a switch argument to control whether it’s enabled?

primus/core/utils/tracer.py
```diff
@@ -0,0 +1,227 @@
...
+        if not self.enabled:
+            return
+
+        self.rank_span = self.tracer.start_span(
+            name=f"rank_{self.rank}",
+            context=Context(),
+            kind=SpanKind.INTERNAL,
+            attributes={"rank": self.rank}
+        )
+        print(f"[Tracer Rank {self.rank}] rank trace_id {self.rank_span.get_span_context().trace_id} span_id {self.rank_span.get_span_context().span_id}")
```
wenxie-amd: change print() to the logger, so we can config it with INFO/WARN/DEBUG levels

third_party/Megatron-LM
lhzhang333, line 1: Are these third_party updates made by mistake?

primus/modules/trainer/megatron/trainer.py
haishuok0525: There is an env variable in tracer implementation. This env variable is default set false. Tracer will do nothing when invoked while this env variable is set false.

primus/core/utils/tracer.py
haishuok0525: All print() codes are changed to debug() from primus logger.

PR #115, opened 2025-07-08
[UT] Add deterministic extra check and unit test

RuibinCheung:
* Add extra deterministic check. Currently grouped gemm can't bitwise equal on ROCm.
* Add deterministic unit test for regression test.


wenxie-amd (dismissed):
LGTM

Xiaoming-AMD (approved):
LGTM

PR #116, opened 2025-07-09
fix(trainer): auto-enable tensorboard when profiling is enabled

Xiaoming-AMD:
When --profile is enabled in MegatronTrainer, this change ensures that
TensorBoard logging is automatically turned on (i.e., disable_tensorboard=False),
so profiler data can be tracked correctly.

This avoids confusion where profiling runs produce no TensorBoard output due to
disable_tensorboard being True by default.


wenxie-amd (approved):
LGTM

PR #134, opened 2025-07-22
feat(turbo): fit PrimusTurboAttention with cp

ChengYao-amd:


primus/modules/trainer/megatron/pre_trainer.py
```diff
@@ -14,6 +14,11 @@
 from megatron.training import get_args, get_timers
 from megatron.training.utils import get_batch_on_this_cp_rank, get_batch_on_this_tp_rank
 
+from primus.backends.megatron.core.utils import (
+    produce_attention_sharder,
+    shard_batch_on_this_cp_rank,
+)
+
```
wenxie-amd: If the user hasn't installed Primus Turbo, this import will raise an error. Move the import to the line 45.

primus/modules/trainer/megatron/pre_trainer.py
```diff
@@ -35,7 +40,12 @@ def get_batch(self, data_iterator):
         batch = get_batch_on_this_tp_rank(data_iterator)
 
         # slice batch along sequence dimension for context parallelism
-        batch = get_batch_on_this_cp_rank(batch)
+        args = get_args()
+        if args.context_parallel_size > 1 and args.enable_primus_turbo:
```
wenxie-amd: Shall we add a specific argument for this feature, if user enable primus turbo, all the features from turbo are enabled, it's hard to debug.

ChengYao-amd: I meet the same issue when profiling, currently all features in turbo are controlled by one flag. Shall we add a config yaml file to control the fine-grained feature whether to be enabled?

wenxie-amd (approved):
LGTM

PR #153, opened 2025-08-13
feat: reduce cpu sync of moe_router_force_load_balancing

RuibinCheung:
* Reduce cpu sync of moe_router_force_load_balancing


wenxie-amd (approved):
LGTM

PR #166, opened 2025-08-22
feat(router): moe_router_force_load_balancing support flexdispatcher

zhenhuang12:


RuibinCheung:
The NV's implementation is not totally balance. We need to measure the performance gap between their implementation and our. Please take a look. @wenxie-amd

PR #178, opened 2025-09-05
（feat)async tp: adapt async-tp for te2.x api

llying-001:
(1) adapt for async tp for te2.x api except ub_bulk_overlap(fp8)
(2) update async tp test for backward

### DIFF
--- a/primus/backends/transformer_engine/pytorch/cpp_extensions/gemm.py
+++ b/primus/backends/transformer_engine/pytorch/cpp_extensions/gemm.py
@@ -1,5 +1,6 @@
 ###############################################################################
-# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Modification Copyright© 2025 Advanced Micro Devices, Inc. All rights reserved.
 #
 # See LICENSE for license information.
 ###############################################################################
--- a/primus/backends/transformer_engine/pytorch/module/base.py
+++ b/primus/backends/transformer_engine/pytorch/module/base.py
@@ -1,8 +1,10 @@
 ###############################################################################
-# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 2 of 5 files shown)


primus/backends/transformer_engine/transformer_engine_torch/gemm.py
```diff
@@ -0,0 +1,309 @@
...
+            elif comm_type == ptex.CommOverlapType.RS:
+                fn = comm_overlap.split_overlap_rs
+                assert extra_output is not None, "split_overlap_rs requires extra output"
+                args = tuple(args + (extra_output,))
+            else:
+                raise ValueError(
+                    f"TP comm overlap on, but provided {bulk_overlap=} and {comm_type=} are invalid"
+                )
+            fn(*args)
+        else:
```
zhenhuang12, line 301: directly call tex.generic_gemm maybe better when not enable comm_overlap

zhenhuang12 (dismissed):

primus/backends/transformer_engine/transformer_engine_torch/gemm.py
llying-001: If the function has already been replaced in the patch, wouldn't that cause a recursive call issue?

primus/backends/transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -5,219 +5,372 @@
 ###############################################################################
```
wenxie-amd, line 6: Please also add NVIDIA’s license, since this is partly derived from TE.

primus/backends/transformer_engine/pytorch/module/base.py
```diff
@@ -7,6 +7,7 @@
 from typing import Optional, Union
```
wenxie-amd, line 9: Please also add NVIDIA’s license, since this is partly derived from TE.

primus/backends/transformer_engine/transformer_engine_torch/gemm.py
```diff
@@ -0,0 +1,309 @@
+from megatron.core.utils import is_te_min_version
```
wenxie-amd, line 9: add license too

wenxie-amd (approved):
LGTM

PR #195, opened 2025-09-16
feat: align primus-turbo fp8 linear's args to megatron

RuibinCheung:
## Context
Primus-Turbo float8 gemm flag is not support control by megatron's fp8 arguments (e.g fp8 recipe...). This PR convert megatron fp8 recipe to Primus-Turbo's quant configuration. Then we can control fp8 gemm in turbo through megatron arguments.

## Main changes
1. Patch `FP8GlobalStateManager` to support primus-turbo if the module from Primus-Turbo it can get primus-turbo quant config from context manager. It's not breaking change and it's compatible with TE.
2. Add feature: specify some fp8 linear layer to bf16. 
3. Remove `enable_turbo_gemm_float8` because we should control fp8 gemm by megatron's fp8 arguments.
4. Simplify arguments.

Related Primus-Turbo's PR: https://github.com/AMD-AGI/Primus-Turbo/pull/106

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,6 +41,7 @@ jobs:
       - uses: isort/isort-action@v1
         if: always()
         with:
+          isort-version: "5.13.2"
           configuration: "--profile black --check-only --diff"
       - uses: psf/black@stable
         if: always()


primus/configs/modules/megatron/trainer_base.yaml
```diff
@@ -57,6 +57,10 @@ fp8_param_gather: false
 te_rng_tracker: false
 inference_rng_tracker: false
 
+first_last_layers_bf16: false
+num_layers_at_start_in_bf16: 0
```
wenxie-amd: The default value for num_layers_at_start_in_bf16 is 1 (arguments.py), is there any particular reason it was changed to 0?

RuibinCheung: ACK. It should be align to default value. Thx ~

wenxie-amd (approved):
LGTM

wenxie-amd (approved):

PR #244, opened 2025-10-20
Support Megatron's a2a and deepep overlap in pipeline

yuankaichen-amd:
This is just an interface change to enable the overlap code to run.

Functionally speaking:

1. a2a overlap works as intended when pp_warmup=False;
2. deepep overlap doesn't work -- 
  2.1  GPU hangs when turbo_deepep_use_comm_stream=True, pp_warmup=False;
  2.2  turbo_deepep_use_comm_stream=False, pp_warmup=False, was able to run 1 iteration and then got nan error;
  2.3  turbo_deepep_use_comm_stream=False, pp_warmup=True,  was able to run, but no overlaps.

We will need to follow up on the issues above, but the code change in this PR is correct and safe, unrelated to the issues.


wenxie-amd (approved):
LGTM

PR #245, opened 2025-10-21
fix: parse moe_layer_freq from string to list in Megatron config adapter

Xiaoming-AMD:
This PR fixes a parsing issue where moe_layer_freq was being interpreted as a string (e.g., "([0]*1+[1]*26)") instead of a valid Python list or integer when loading Megatron configs.
The adapter now automatically converts string expressions to lists using ast.literal_eval, ensuring compatibility with YAML and CLI configurations.


wenxie-amd (approved):
LGTM

PR #256, opened 2025-10-24
test(torchtitan): add model unit tests for TorchTitan backend

Xiaoming-AMD:
This PR introduces initial unit tests for the TorchTitan backend in the Primus framework. These tests help verify model construction, trainer instantiation, and config overrides — forming the basis of a robust regression suite for Titan support.

---


wenxie-amd (approved):
LGTM

PR #272, opened 2025-11-06
[Fix] import MXLinear from Primus Turbo

GeneDer:
RE: https://github.com/AMD-AGI/Primus-Turbo/pull/145/files#diff-11d1092259d933e0044a34ea1b901386b9f344e91bc185755e019ccc9bba4aa0L14-R14

Primus Turbo recently renamed this `MXLinear` to `Float8Linear`. Update it in Primus accordingly to make them compatible.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ on:
   pull_request:
 
 env:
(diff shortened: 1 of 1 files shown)


wenxie-amd:
The mxfp8 feature is still under development, don't use it yet.

PR #286, opened 2025-11-17
specify recompute_layer_ids

llying-001:
Specify the exact IDs of layers to recompute, enabling more flexible memory reduction                                                                                                                                                                                                                                                                                                      Using `recompute_layer_ids=[layer_id_0, layer_id_1,...]` together with `recompute_granularity=full`, where layer_id ranges from 0 to num_layers - 1.


wenxie-amd (approved):
LGTM

PR #294, opened 2025-11-18
fix errors when patch_moe_overlap is set and distribute Wgrads to overlap with different comms

yuankaichen-amd:


yuankaichen-amd:
I also included an updated overlap implementation where 

(1) WgradStore allows flushing a given number of items (previously it always flushes all items);
(2) Wgrads from GroupedGemm (2x #layers) are now distributed to overlap with different types communication (1 with pp, others with dispatch or combine).

PR #298, opened 2025-11-20
feat(config): make Megatron workspace path configurable via env var

poznano-amd:
In order to facilitate automated benchmarking & profiling, we'd like the option to be able to configure where the workspace gets saved into the wider filesystem.  

This PR adds the ability to control the base workspace path via a new env var:  `PRIMUS_WORKSPACE`.
* When set, all of the resulting workspace directory structure is dumped to the specified location
* When unset, the default remains as `./output/` (same as the original setting)

We have successfully tested the `PRIMUS_WORKSPACE` env var in our automation environment for profiling workloads.


wenxie-amd:
The CI (build-docker, run-unittest) failed. It seems you forked Primus, but the CI requires repository secrets (like HF_HOME, PRIMUS_DOCKER_HUB_TOKEN), and these secrets don't exist in your repository.

PR #315, opened 2025-11-28
backend(maxtext): support custom model condifg and update model args via cli

llying-001:
reconstruct maxtext model config:
(1) support custom model condifg like grok1.yaml by setting model_name="default" and specify model args
(2) support to update model args via cli like test_maxtext_trainer.py
(3) add grok1 model config which needs to update and more test cases for maxtext trainer

### DIFF
--- a/primus/modules/trainer/maxtext/pre_trainer.py
+++ b/primus/modules/trainer/maxtext/pre_trainer.py
@@ -71,7 +71,7 @@ class MaxTextPretrainTrainer(BaseModule):
         flat_overrides = {}
         for k, v in override_args.items():
             if k != "model":
-                raise ValueError(f"MaxText Pre-Trainer: Nested dicts are not supported for override key {k}.")
+                raise ValueError(f"Only the 'model' key is supported for overrides, found: {k}")
             if not isinstance(v, dict):
                 raise ValueError(
                     f"MaxText Pre-Trainer: The value for 'model' must be a dict, got {type(v).__name__}."


primus/configs/models/maxtext/grok1.yaml
```diff
@@ -0,0 +1,22 @@
+bases:
+  - model_base.yaml
```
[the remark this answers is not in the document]
llying-001: grok1 model is not in native maxtext's model list. Only by setting model_name="default", we can specify model configs.

primus/configs/models/maxtext/model_base.yaml
```diff
@@ -0,0 +1,114 @@
+model_name: "default"
+override_model_config: False # When set to true allows overriding model parameters via CLI for the purpose of debugging/testing.
```
Xiaoming-AMD, line 2: Don't need a separate argument to control whether model-parameter overrides are enabled. As long as --model.** is set, Primus should attempt to override the corresponding parameter, and if the parameter does not exist, it should raise an error

primus/configs/models/maxtext/deepseek_v2_16B.yaml
```diff
@@ -1,2 +1,7 @@
+bases:
```
Xiaoming-AMD, line 1: According to the latest main branch, the parsing here uses the extends mechanism and no longer supports base or include. Refer to PR https://github.com/AMD-AGI/Primus/pull/312
. Also, if the values defined in base were default values, you no longer need to specify them.

examples/maxtext/configs/MI300X/grok1-pretrain.yaml
```diff
@@ -0,0 +1,62 @@
...
+      base_output_directory: "./output"
+      steps: 50
+      log_period: 10
+
+      # data
+      dataset_type: "synthetic"
+      hf_access_token: ${HF_TOKEN:""}
+
+      # checkpoint
+      enable_checkpointing: False
```
Xiaoming-AMD, line 24: Use false/true consistently.

PR #584, opened 2026-03-09
feat(megatron): add pp_dump_data patches for PP schedule visualization

HuangWei-95:
- Add before_train patch to wrap get_forward_backward_func with schedule_wrapper
- Primus pipeline / ZeroBubble: only wrap schedule; handlers already use fwd_bwd_wrapper
- Megatron native: wrap schedule and patch forward_step/backward_step via set_dump_pp_data_patch
- Add guarded schedule_wrapper to avoid double-wrap in legacy MegatronTrainer flow
- Add after_train patch to call dump_pp_data for core runtime


HuangWei-95:
./runner/primus-cli direct --log_file logs/llama2_7B-BF16.log  -- train pretrain --config examples/megatron/configs/MI300X/llama2_7B-BF16-pretrain.yaml  --train_iters 10 --pipeline_model_parallel_size 4 --global_batch_size 64 --dump_pp_data true

image

PR #589, opened 2026-03-10
refactor(config): route Megatron override compatibility through launcher policy

WangLingxun:
- Declare the unknown CLI override passthrough in the Megatron module preset so the compatibility rule stays with the preset it applies to.
- Read the generic launcher policy during parsing and preserve it through runtime normalization so legacy Megatron overrides remain compatible without backend-specific branching.


primus/core/launcher/parser.py
```diff
@@ -135,7 +144,9 @@ def parse_args(extra_args_provider=None, ignore_unknown_args=False):
 
     overrides = parse_cli_overrides(unknown_args, type_mode="legacy")
     pre_trainer_cfg = primus_config.get_module_config("pre_trainer")
-    _check_keys_exist(pre_trainer_cfg, overrides)
+    # For megatron, allow passthrough so YAML key cleanup does not break CLI overrides.
+    if not _allow_backend_override_passthrough(pre_trainer_cfg, "megatron"):
```
Xiaoming-AMD: Backend-specific logic should not appear here.

primus/core/backend/backend_registry.py
```diff
@@ -48,6 +48,13 @@ class BackendRegistry:
     # Backend → list of setup hooks
     _setup_hooks: Dict[str, List[Callable]] = {}
 
+    # Backends that may accept CLI overrides not declared in module YAML presets.
+    _backend_unknown_cli_overrides = {
```
Xiaoming-AMD: backend-special logic should not appear here（core/backend/* ）

PR #682, opened 2026-04-22
Lazy-import matplotlib/markdown2/weasyprint in preflight

yeandy:
Move optional dependencies to lazy imports so preflight can run with only torch installed. matplotlib is now imported only when --plot is used; markdown2 and weasyprint are imported only inside md_to_pdf() (skipped when --disable-pdf is passed).


amd-ama10002-2:
Why was this closed? I thought we had a customer requesting that we reduce the number of dependencies for preflight checks? 

Also, I guess this is a public repo, but we have an internal JIRA ticket tracking some customer requests for feature enhancements, You can message me on Teams if you want the link 🙂🙏

yeandy:
Closed in favor of a fresh one https://github.com/AMD-AGI/Primus/pull/686

PR #706, opened 2026-04-29
feat(preflight): add --no-split-nodes-subgroup to skip subgroup inter-node tests

yeandy:
…-node tests

Allows skipping 2-node and 4-node subgroup inter-node comm tests, running only the all-node allreduce/alltoall and ring p2p.


amd-ama10002-2:
Ah I wasn't fast enough at reviewing it but FWIW, it LGTM 👍 I was gonna approve it anyways 😅

PR #710, opened 2026-04-30
fix(preflight): add barrier + delay between comm group destroy/create…

yeandy:
### Problem

At 128-node scale (1024 ranks), the preflight perf tests intermittently fail with:

```
torch.distributed.DistNetworkError: The server socket has failed to listen on any local network address. ... Address already in use
```

This happens because the preflight tool rapidly destroys and recreates `torch.distributed` process groups between test phases (intra-node → inter-node → P2P → ring). When communicators are destroyed, the underlying RCCL/NCCL sockets enter `TIME_WAIT` state. If the next phase tries to bind the same port before the OS releases it, the bind fails.

### Fix

Adds `barrier_after_comm_destroy()` — a global barrier + configurable sleep after every `dist.destroy_process_group()` call in the preflight tool:

1. **Barrier** — ensures all ranks have finished teardown before any rank starts the next phase (prevents fast ranks from racing ahead)
2. **Sleep** (default 2s) — gives the OS time to release sockets from `TIME_WAIT`

#### New CLI flag

```
--comm-cleanup-delay-sec <float>   (default: 2.0)
```

- Set to `0` to disable the sleep (keeps the barrier only)
- Increase to `5` if EADDRINUSE still occurs at very large scale

#### Files changed

- `utility.py` — new `barrier_after_comm_destroy()` helper with logging
- `preflight_perf_test.py` — inter-test barriers between phases
- `preflight_args.py` — `--comm-cleanup-delay-sec` argument
- `intra_node_comm.py`, `inter_node_comm.py`, `inter_node_comm_p2p.py`, `inter_node_ring_p2p.py` — call barrier after destroying sub-groups
- `docs/preflight.md`, `docs/preflight-direct.md` — document new flag + troubleshooting

### Testing

| Run | Scale | Result |
|-----|-------|--------|
| preflight end-to-end | 4N / 32 GPUs | **PASSED** — all phases, zero EADDRINUSE |
| preflight end-to-end | 6N / 48 GPUs | **PASSED** — all phases, zero EADDRINUSE |
| stress test (50 cycles group create/destroy) | 10N / 80 GPUs | **PASSED** with and without barrier |

> **Note:** Could not validate at 128N due to not having that same number of nodes. Recommend customer validation at target scale. If EADDRINUSE persists, increase delay with `--comm-cleanup-delay-sec 5`.

### Overhead

~10s total added to a full preflight run (5 barrier transitions × 2s each). Negligible for a health-check tool.

### DIFF
--- a/docs/preflight-direct.md
+++ b/docs/preflight-direct.md
@@ -192,6 +192,7 @@ See [Preflight](./preflight.md) for the full list. The most common are:
 - Selection: `--host`, `--gpu`, `--network`, `--perf-test`
 - Reporting: `--dump-path`, `--report-file-name`, `--disable-pdf`
 - Perf-test extras: `--plot`
+- Reliability: `--comm-cleanup-delay-sec <float>`
 
 If you do not pass `--report-file-name`, the wrapper auto-generates a unique one of the form:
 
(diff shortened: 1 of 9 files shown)


amd-fuyuajin (approved):
I did a quick test on 4N MI355X cluster and it worked as expected. I don't have more nodes to test.

yeandy:
> I did a quick test on 4N MI355X cluster and it worked as expected. I don't have more nodes to test.

Thanks! I think we can merge for now to allow users to test at larger scale.

PR #720, opened 2026-05-09
Add Slurm preflight bisection helper

amd-ama10002-2:
## Summary

This PR adds a Slurm-based preflight bisection helper script for isolating suspect nodes when multi-node preflight perf tests fail or hang. The new workflow runs preflight on recursively smaller node subsets, records per-trial logs, and writes a final summary of suspected faulty nodes.

It also adds two pytests for validating the bisection using a real pre-flight check in a slurm allocation
1. Test that the bisection passes on a set of healthy nodes in a slurm allocation
2. Test that bisection can correctly identify an unhealthy node (by temporarily creating a "bad node" on during execution)

in collaboration with: @akasharidas 

## Motivation

To reduce that manual effort when debugging NCCL/RCCL hangs or node-specific failures when using our preflight checks

## Technical Details

- Added `tools/preflight_bisect/bisect.py`, which expands a Slurm nodelist, runs `runner/run_preflight_direct.sh --perf-test` on subsets via `srun`, and recursively bisects failing or hanging subsets until suspect singleton nodes are identified.
- Added `tools/preflight_bisect/fake_runner.sh`, a drop-in test runner that simulates one bad node on a live Slurm allocation without requiring an actually broken host.
- Added `tests/unit_tests/tools/test_preflight_bisect_slurm.py` for opt-in live-Slurm integration coverage. The tests can derive `BISECT_NODELIST` from `SLURM_NODELIST`, omit `--partition` when no partition is known, and default the fake bad node to the last resolved hostname.
- Updated `docs/preflight-direct.md` with the automated node-bisection workflow and usage example.

## Test Plan

- Run the pytest integration test file in a real slurm cluster:

```bash
(preflight) andrew.ma@amd.com@tus1-vm-amd-prj2-k8s-010:~/Primus$ python3 -m pytest \
  tests/unit_tests/tools/test_preflight_bisect_slurm.py \                       
  -v \
  --basetemp=/tmp/preflight-bisect-pytest
```
  
## Test Result
- Test passes and result appears as expected
- The test that a set of healthy nodes can pass takes 3min on 4N and it can take longer on larger sets of nodes. 

<details><Summary> Screenshot of passing test </Summary> 

Screenshot 2026-05-08 190640

File output looks like the expected output 1 test with all passing and one test that identifies the last node as a bad node

Screenshot 2026-05-08 190609

</details> 

@akasharidas did further testing on 16 nodes and reports this:

> I managed to run it on 16N. With the simulated "bad node" failure, the preflight check fails very quickly, so it goes through the entire bisection very quickly. With a simulated hang (rather than failure), the bisection might take longer, since it will wait until the preflight timeout completes before deciding that its hung. 

<details> <Summary> 16N Test results catching a simulated hang </Summary>

```bash 
(preflight) akash.haridas@amd.com@tus1-p3-g38:~/Primus$ python tools/preflight_bisect/bisect.py     --nodelist "$SLURM_NODELIST"     --partition amd-tw --reservation DO_NOT_REMOVE_TANVEERM     --output-dir ~/Primus/output/bisect-$(date +%Y%m%d-%H%M%S)     --trial-timeout-sec 600     --slurm-time 00:15:00     --runner tools/preflight_bisect/fake_runner.sh  --preflight-env USING_AINIC=1     --preflight-env NCCL_IB_GID_INDEX=3     --preflight-env NCCL_CROSS_NIC=0     --preflight-env NCCL_PXN_DISABLE=0     --preflight-env BAD_NODE=tus1-p3-g2 2>&1 | tee ~/Primus/output/bisect-latest.log
[trial 000] N=16 tus1-p3-g2..tus1-p3-g59 -> trial-000.log
[trial 001] N=8 tus1-p3-g50..tus1-p3-g59 -> trial-001.log
[trial 002] N=8 tus1-p3-g2..tus1-p3-g32 -> trial-002.log
[trial 003] N=4 tus1-p3-g26..tus1-p3-g32 -> trial-003.log
[trial 004] N=4 tus1-p3-g2..tus1-p3-g25 -> trial-004.log
[trial 005] N=2 tus1-p3-g15..tus1-p3-g25 -> trial-005.log
[trial 006] N=2 tus1-p3-g2..tus1-p3-g14 -> trial-006.log
[trial 007] N=1 tus1-p3-g14 -> trial-007.log
[trial 008] N=1 tus1-p3-g2 -> trial-008.log
2026-05-11T17:56:04.151338+00:00 bisect nodes=16
[000] N=16 FAIL  nodes=tus1-p3-g2..tus1-p3-g59
[001] N= 8 PASS  nodes=tus1-p3-g50..tus1-p3-g59
[002] N= 8 FAIL  nodes=tus1-p3-g2..tus1-p3-g32
[003] N= 4 PASS  nodes=tus1-p3-g26..tus1-p3-g32
[004] N= 4 FAIL  nodes=tus1-p3-g2..tus1-p3-g25
[005] N= 2 PASS  nodes=tus1-p3-g15..tus1-p3-g25
[006] N= 2 FAIL  nodes=tus1-p3-g2..tus1-p3-g14
[007] N= 1 PASS  nodes=tus1-p3-g14
[008] N= 1 FAIL  nodes=tus1-p3-g2
SUSPECT_NODES: tus1-p3-g2
```
</details>

Follow-up testing with simulated hangs also looked good on 16N. One caveat: `--trial-timeout-sec` should be set high enough for a healthy subset to finish, otherwise slow-but-good trials can be counted as false failures and cause the bisection to explore extra paths before converging.


tools/preflight_bisect/bisect.py
```diff
@@ -0,0 +1,326 @@
...
+    ]
+    if args.partition:
+        cmd.extend(["-p", args.partition])
+    # The new wrapper run_preflight_direct.sh does NOT accept "--env KEY=VALUE".
+    # Per-trial env overrides are propagated via srun --export. ALL keeps the
+    # caller's environment (notably VENV_ACTIVATE) intact; the trailing K=V
+    # pairs override / add on top. SLURM tokenizes --export on commas, so
+    # values must not contain ',' or whitespace (NCCL flags never do).
+    if args.preflight_env:
+        cmd.append("--export=ALL," + ",".join(args.preflight_env))
```
yeandy: When args.preflight_env is empty, no --export flag is passed to srun at all. On clusters where PropagateEnvironment=no is set in slurm.conf, this silently loses VENV_ACTIVATE and other required env vars, causing the runner to fail with a confusing error.

Consider always passing --export=ALL and only appending the extra K=V pairs when present:
```suggestion
    export_val = "ALL"
    if args.preflight_env:
        export_val += "," + ",".join(args.preflight_env)
    cmd.append(f"--export={export_val}")
```

tools/preflight_bisect/bisect.py
```diff
@@ -0,0 +1,326 @@
...
+    for t in sorted(trials, key=lambda trial: trial["idx"]):
+        nlist = t["nodes"]
+        r = _format_node_range(nlist)
+        lines.append(f"[{t['idx']:03d}] N={t['n']:2d} {t['status'].upper():4s}  nodes={r}")
+    if suspects:
+        lines.append("SUSPECT_NODES: " + " ".join(sorted(set(suspects))))
+    else:
+        lines.append("SUSPECT_NODES: (none)")
+    path.write_text("\n".join(lines) + "\n", encoding="utf-8")
+    print(path.read_text(), end="")
```
yeandy: This way you don't have to read back what you wrote 
```suggestion
    summary = "\n".join(lines) + "\n"
    path.write_text(summary, encoding="utf-8")
    print(summary, end="")
```

tests/unit_tests/tools/test_preflight_bisect_slurm.py
```diff
@@ -0,0 +1,218 @@
...
+    cmd = [
+        sys.executable,
+        str(BISECT_PY),
+        "--nodelist", env["BISECT_NODELIST"],
+        "--output-dir", str(tmp_path),
+        *preflight_env_args,
+        *extra_args,
+    ]
+    if env.get("BISECT_PARTITION"):
+        cmd[4:4] = ["--partition", env["BISECT_PARTITION"]]
```
yeandy: cmd[4:4] is fragile -- if anyone adds or reorders args in the cmd list above, this silently inserts --partition in the wrong position. Building the command with conditionals avoids the coupling to a magic index:
```suggestion
    cmd = [
        sys.executable,
        str(BISECT_PY),
        "--nodelist", env["BISECT_NODELIST"],
        "--output-dir", str(tmp_path),
    ]
    if env.get("BISECT_PARTITION"):
        cmd.extend(["--partition", env["BISECT_PARTITION"]])
    cmd.extend([*preflight_env_args, *extra_args])
```

tests/unit_tests/tools/test_preflight_bisect_slurm.py
```diff
@@ -0,0 +1,218 @@
...
+        timeout=timeout,
+    )
+    summary_path = tmp_path / "summary.txt"
+    summary_text = summary_path.read_text(encoding="utf-8") if summary_path.exists() else ""
+
+    if result.returncode != 0 and not summary_text:
+        pytest.fail(
+            f"bisect.py exited {result.returncode} and wrote no summary.\n"
+            f"stdout:\n{result.stdout}\nstderr:\n{result.stderr}"
+        )
```
yeandy: bisect.py always returns 0 on normal completion (even when suspects are found), so returncode != 0 unconditionally means a crash or validation error. The and not summary_text guard lets partial-crash scenarios (where the code managed to write a partial summary_text) slip through. This should just be:
```suggestion
    if result.returncode != 0:
        pytest.fail(
            f"bisect.py exited {result.returncode} and wrote no summary.\n"
            f"stdout:\n{result.stdout}\nstderr:\n{result.stderr}"
        )
```

tests/unit_tests/tools/test_preflight_bisect_slurm.py
```diff
@@ -0,0 +1,218 @@
...
+# use SLURM_NODELIST from the current allocation.
+#
+# Running from inside a Slurm allocation
+# --------------------------
+#   cd ~/Primus
+#   export VENV_ACTIVATE=~/envs/preflight/.venv/bin/activate
+#   python3 -m pytest tests/unit_tests/tools/test_preflight_bisect_slurm.py -v
+
+
+
```
yeandy: No need for these extra lines
```suggestion
```

tests/unit_tests/tools/test_preflight_bisect_slurm.py
```diff
@@ -0,0 +1,218 @@
+###############################################################################
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+#
+# See LICENSE for license information.
+###############################################################################
+#
+# Integration tests for tools/preflight_bisect/bisect.py. These require a live
```
yeandy, line 7: The file header correctly says "Integration tests," but the file lives under tests/unit_tests/. Should we rename to `tests/integration_tests`?

tools/preflight_bisect/bisect.py
```diff
@@ -0,0 +1,326 @@
...
+            )
+            try:
+                rc = proc.wait(timeout=args.trial_timeout_sec)
+                return "pass" if rc == 0 else "fail"
+            except subprocess.TimeoutExpired:
+                try:
+                    os.killpg(proc.pid, signal.SIGKILL)
+                except ProcessLookupError:
+                    pass
+                proc.wait(timeout=60)
```
yeandy: If the SIGKILL'd process is stuck in uninterruptible I/O (e.g., hung NFS mount -- not unusual on shared HPC filesystems), proc.wait(timeout=60) raises subprocess.TimeoutExpired, which is unhandled and would crash the entire bisection without writing a summary.
```suggestion
                try:
                    proc.wait(timeout=60)
                except subprocess.TimeoutExpired:
                    pass  # process is stuck in D-state; SIGKILL is pending, move on
```

tools/preflight_bisect/bisect.py
yeandy, line 1: nit: Missing the AMD copyright/license header that the other new files in this PR (and the rest of the repo) include. Same applies to fake_runner.sh.

tests/unit_tests/tools/test_preflight_bisect_slurm.py
akasharidas: Will defer to @amd-ama10002-2 on this.

amd-ama10002-2: Good callout 🙏

IMHO, I think we should leave it as. It's true that this is an integration test that is under a directory named `unit_tests/` however the `unit_tests/` directory contains a lot of other real unit tests, so I don't want to change the directory name. 

I also don't want to create an `integration_tests/` directory for one file today, but maybe in the future if we get more integration tests then we can have another PR to reorganize the pytest files in the repo? 

(I'm also imagining that if they package this work later, then they might use a source folder and a tests folder directory structure. In that case, they might even just get rid of the unit tests subdirectory and just have the test folder be a mirror structure of the source code, so I don't want to reorganize the directory at this time. Maybe that can be a future PR )

PR #734, opened 2026-05-27
Cluster Perf tests - EP benchmarking & RDMA Perf and Cluster Env mapping recommendation tool

lcskrishna:
This PR enables the following perf tests and tools into Primus. 
Here are the summary of perf tests and tools added.

* Large EP Performance tests (MoRI-EP) - ``` benchmark/kernel/ep_bench``` - Used for Microbenchmarking Large Expert Parallelism using MoRI
* RDMA Perf (IB_write) tests - ``` benchmark/kernel/rdma_perf``` - Used to validate the NIC performance of a cluster between two nodes. 
* Cluster RDMA Env mapping Tool - ``` tools/cluster-rdma-env-recommender ``` - Details of cluster like Firmware, GID, RDMA -> NETDEV mapping and NIC vendor and few other recommendations. 

Each of these perf tests and tools have their respective README.

### DIFF
--- a/benchmark/kernel/ep_bench/README.md
+++ b/benchmark/kernel/ep_bench/README.md
@@ -1,0 +1,163 @@
+# MoRI EP-Bench (Slurm, no torchrun)
+
+This directory provides a Slurm-based launcher for the **MoRI** expert-parallel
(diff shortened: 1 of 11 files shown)


lcskrishna:
cc: @alfuyao-amd

PR #761, opened 2026-06-10
[Fix] destory pg for megatorn trainer on cleanup

GeneDer:


GeneDer:
This doesn't work. The process still hangs

PR #766, opened 2026-06-12
Dev/update readme pip install

wenxie-amd:


amd-fuyuajin:
Add one testing point here. I tested the `setup.sh` script on MI355X. Setting up the whole training environment took quite a few hours, because several packages were built from source. This info has been specified in the instruction.
After installing the packages on the host, I tested llama3.1-8B model and Qwen3.5-35B-A3B model. Both could run.

PR #769, opened 2026-06-16
refactor(turbo): rename deprecated turbo flags and hard-assert on rem…

wenxie-amd:
## Summary

Removes two deprecated Primus-Turbo flags and replaces them with their canonical names across the entire repo:

- `use_turbo_grouped_mlp` → **`use_turbo_grouped_gemm`**
- `use_turbo_parallel_linear` → **`use_turbo_gemm`**

Previously these deprecated flags only emitted a warning and were silently OR'd into the new flags inside `validate_args_on_rocm`, but the value was **never written back to `args`**. As a result, setting `use_turbo_grouped_mlp=True` (or `use_turbo_parallel_linear=True`) did **not** actually activate the Primus-Turbo grouped-GEMM / dense-GEMM kernel paths — the spec provider keys off `use_turbo_grouped_gemm` / `use_turbo_gemm`, which stayed `False`. This silently masked the turbo kernels and made A/B comparisons misleading.

This PR makes the deprecated flags hard errors so the misconfiguration can't slip through unnoticed.

## Changes

- **Configs**: renamed the flags in all `examples/**` and `tests/**` YAMLs. Dropped the deprecated keys entirely from the base `primus/configs/modules/megatron/primus_turbo.yaml` (kept only the canonical `use_turbo_gemm` / `use_turbo_grouped_gemm`).
- **Validation** (`primus/modules/trainer/megatron/utils.py`): replaced the deprecation warning + silent aliasing with a hard `assert` that points users to the new flag when either removed flag is set.
- **Spec provider** (`transformer_engine_spec_provider.py`): removed the now-dead `or use_turbo_parallel_linear` fallbacks in the linear module selection.
- **Patches** (`te_spec_provider_patches.py`, `legacy_grouped_mlp_wgrad_patches.py`): removed the dead `use_turbo_grouped_mlp` getattr fallbacks in patch conditions.
- **Projection** (`training_config.py`, `performance_projection/projection.py`): renamed the `ModelConfig` field and the auto-enable logic. This also fixes a latent bug where the projection's `ModelConfig.use_turbo_grouped_mlp` field name no longer matched the runtime arg, so the turbo grouped-GEMM performance model was never selected.
- **Examples & tests**: aligned `examples/moe_package/**`, `examples/customer_package/**` launch scripts and the projection profiler unit test to the new flag names.

## Behavior change

Setting `use_turbo_grouped_mlp` or `use_turbo_parallel_linear` now raises:

use_turbo_grouped_mlp has been removed; please use use_turbo_grouped_gemm instead. use_turbo_parallel_linear has been removed; please use use_turbo_gemm instead.

Existing configs that relied on the deprecated names must migrate to the new flags.

### DIFF
--- a/examples/customer_package/run_qwen3_235b_a22b_pretrain_mi355x.sh
+++ b/examples/customer_package/run_qwen3_235b_a22b_pretrain_mi355x.sh
@@ -108,7 +108,7 @@ for feature in "${MoE_Features[@]}"; do
         ;;
     2)
         ensure_primus_turbo
-        FEATURE_ARGS+=("--use_turbo_grouped_mlp" "True")
+        FEATURE_ARGS+=("--use_turbo_grouped_gemm" "True")
         ;;
     3)
         FEATURE_ARGS+=("--cross_entropy_fusion_impl" "te")
--- a/examples/customer_package/run_qwen3_30b_a3b_pretrain_mi355x.sh
+++ b/examples/customer_package/run_qwen3_30b_a3b_pretrain_mi355x.sh
@@ -102,7 +102,7 @@ for feature in "${MoE_Features[@]}"; do
         ;;
     2)
         ensure_primus_turbo
-        FEATURE_ARGS+=("--use_turbo_grouped_mlp" "True")
+        FEATURE_ARGS+=("--use_turbo_grouped_gemm" "True")
         ;;
     3)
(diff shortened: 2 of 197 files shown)


RuibinCheung (approved):
LGTM

Xiaoming-AMD (approved):

PR #806, opened 2026-06-28
ci: bump Primus-Turbo/AITER pins for Flux diffusion (mxfp4) + skip CI on draft PRs

jasainio:
Root of an 18-PR series splitting the Flux diffusion-training feature (training Flux, a DiT text-to-image diffusion model, on Primus/Megatron) out of one large branch for reviewability. This PR targets `main` and is the **first to merge**. The content branches are cut from this branch (`feat/flux/ci-env`), not raw `main`, so they carry the bumped pins + draft guard on their own heads — which lets the rest of the stack open as drafts immediately, before this PR merges.

## What this changes
Updates the public-CI dependency pins to the versions the Flux diffusion feature needs:
- Bumps `PRIMUS_TURBO_COMMIT` to a recent Primus-Turbo `main` build that exposes the mxfp4 `gemm_fp4_impl(..., preshuffled=...)` fast path, and `PRIMUS_TURBO_AITER_COMMIT` to AITER `v0.1.14.post1`. Mirrors the same turbo commit into `benchmark.yaml` (the docker build picks both up through `ci.yaml`'s build-args).
- Adds a **draft-skip guard** to the expensive jobs (`build-docker`, `run-unittest-torch`, `run-unittest-jax`) so the rest of the stack can be opened as Draft PRs without firing the full docker-build + GPU pipeline on every open/sync; `code-lint` stays ungated. Two parts: add `ready_for_review` to the `pull_request` trigger `types` (the default `opened/synchronize/reopened` set omits it, so without this the draft→ready flip triggers nothing) **and** add `if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.draft }}` to the three heavy jobs.

## Why it merges first (but need not block opening)
On the current `main` pins the mxfp4 path fails (`gemm_fp4_impl` rejects the `preshuffled` arg), so content PRs can only go green if their head branch carries the new pins — which stacking on `feat/flux/ci-env` provides without a merge. Merging this first matters at *merge* time: once it's on `main`, children auto-retargeted to `main` inherit the fix and the guard.

No `MEGATRON_PATH` change is needed: the diffusion unit tests run in-process and `tests/conftest.py` already puts the `submodules: recursive`-checked-out `third_party/Megatron-LM` on `sys.path` when `MEGATRON_PATH` is unset.

## Dependencies
None — this is the root prerequisite.

## Test plan
No unit tests of its own. After the runner rebuilds the image, confirm a smoke run is green: `pytest tests/unit_tests/backends/megatron/diffusion -k mxfp4`. The guard can be sanity-checked by opening a throwaway draft PR (expect only `code-lint`) and marking it ready (expect the full pipeline).

## Files
2 (`.github/workflows/ci.yaml` — pins + draft guard; `.github/workflows/benchmark.yaml` — mirrored turbo pin).


luiza-amd:
Update: Primus-Turbo/AITER pins were aligned with main branch and tests pass

RuibinCheung (approved):
LGTM

Xiaoming-AMD (approved):

PR #812, opened 2026-06-28
feat(flux): diffusion data pipeline (energon/synthetic providers, encoders)

jasainio:
Part of an 18-PR series splitting the Flux diffusion-training feature (training Flux, a DiT text-to-image diffusion model, on Primus/Megatron) out of one large branch for reviewability. Targets `feat/flux/flux` — review after it.

## What this changes
The diffusion data layer: dataloader + CUDA prefetch, energon and synthetic dataset providers, the image (VAE) and text (CLIP-L, T5-XXL) encoders, and the image task encoder. Also moves the energon/webdataset pins into `requirements.txt` here, since this layer's source and tests need them at runtime.

## Why it's stacked here
One synthetic-dataset module top-level-imports `flux.utils`, so it bases on `feat/flux/flux` (this also transitively re-parents the prep/trainers/mlperf layers onto the Flux branch).

## Dependencies
Sequenced after the CI-pins PR (`feat/flux/ci-env`); builds on `feat/flux/flux`.

## Test plan
`pytest tests/unit_tests/backends/megatron/diffusion/data` (needs `megatron-energon`/`webdataset` from the moved `requirements.txt`). Validated locally on an AMD GPU container: 50 passed.

## Files
32 (dataloader/prefetch, dataset providers, VAE/CLIP/T5 encoders, task encoder, `requirements.txt` + tests).


luiza-amd:
A short summary of  additional two fixes committed, both in `tests/unit_tests/backends/megatron_bridge/test_megatron_bridge_adapter.py`:

## Fix 1 — `sys_modules_guard` fixture (energon failure)

- **Symptom:** `assert set(stubbed) == set(_BRIDGE_OPTIONAL_PACKAGES)` failed with `Extra items: 'megatron.energon'`.
- **Cause:** `megatron.energon` (installed by this PR) was already cached in `sys.modules`, so the stub installer's `if pkg_name in sys.modules: continue` fast-path skipped it — it never got stubbed.
- **Fix:** The guard fixture now drops the stub-eligible modules **before** the test too (not just after), so each test starts from a clean slate and the forced-missing mock applies to every package.

## Fix 2 — `test_install_transformers_stub_adds_placeholders_for_missing` (transformers failure)

- **Symptom:** `assert 'Glm4vMoeForConditionalGeneration' in []`.
- **Cause:** The test tried to force a real class "missing" via `delattr`, but `transformers`' lazy module re-caches it before the check. On the CI runner's newer `transformers` (which ships all the classes), nothing was seen as missing, so nothing got stubbed → `[]`.
- **Fix:** Append a **synthetic always-missing class** to the real list to drive the "missing" branch deterministically, independent of `transformers` version.

## Common thread

Neither is a bug in the adapter code — both are pre-existing, environment-sensitive test flaws. They only surfaced because editing `requirements.txt` makes CI run the **full** unit-test suite (not a narrow subset), which pulled in this file with the PR's newly installed `megatron-energon` and the runner's newer `transformers`.

wenxie-amd (approved):

PR #830, opened 2026-06-30
chore: bump primus-turbo version

RuibinCheung:
# Description

Bump the pinned Primus-Turbo (and its bundled aiter) revisions to the latest versions, and adapt the Megatron Turbo extensions to the quantization API changes that came with them.

The new Primus-Turbo revision reworks its low-precision quantization APIs:
it adds a `create_quantized_weight` helper, and renames/refactors the grouped segment-M quantization kernel from `quant_fp8_blockwise_segment_m_impl` (col-only, 4 return values) to `quant_fp8_blockwise_segment_m_row_col_impl` (fused row+col, 6 return values).
Without these adaptations the Turbo FP8/FP4 and ZBV split-wgrad paths fail at import time with `ImportError: cannot import name 'quant_fp8_blockwise_segment_m_impl'`, which surfaces as a runtime `LinearWithWeightGradientStore requires primus_turbo` failure in the affected end-to-end tests.

This PR bumps the version pins and updates the Primus side so the Turbo paths build and run against the new revision.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Infra/Build change
- [x] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Bump the pinned `PRIMUS_TURBO_COMMIT` to `a04a233cbfb468dbe21600cbf9db70953428b25c` (primus-turbo #386, "force use nt layout gemm in bwd") and `PRIMUS_TURBO_AITER_COMMIT` to `0f3c58e6edb6754940bcf9fd5f09ccb6f389f52e` (aiter v0.14.0.post1).
- Propagate the version bump across the build/CI/doc surfaces: `.github/workflows/ci.yaml`, `.github/workflows/benchmark.yaml`, `docker/docker-release/Dockerfile.primus-v26.4`, `docs/install-on-host.md`, `tools/installation/setup.sh`, and `tools/daily/safe_wrapper.py`.
- Add a `_maybe_create_quantized_weight_buffers()` helper in `primus/backends/megatron/core/extensions/primus_turbo.py` that prefers the new Primus-Turbo `create_quantized_weight` API (scaling recipe derived from the quant config, per-granularity transpose handled), and falls back to a manual rowwise/colwise quantize on older builds that do not export it yet.
[...]

PR #852, opened 2026-07-06
fix(megatron): correct fp4 quantized-weight buffer kwarg

WangLingxun:
## Summary
Fix a `TypeError` on the first FP4 microbatch in the Turbo `Linear` weight-buffer path.

In `primus/backends/megatron/core/extensions/primus_turbo.py`, the FP4 `Linear` branch called `_maybe_create_quantized_weight_buffers(...)` with a non-existent keyword `need_cache_colwise=`, while the helper's parameter is `disable_parameter_transpose_cache`. Since the helper computes `need_weight_transpose_cache = not disable_parameter_transpose_cache`, the previous code also inverted the intended value.

This was introduced in #830 (primus-turbo version bump), where this one call site added an inverted intermediate `need_cache_colwise = not self.disable_parameter_transpose_cache` instead of passing the flag directly like the other seven call sites.

## Fix
Drop the inverted intermediate and pass `disable_parameter_transpose_cache=self.disable_parameter_transpose_cache`, matching the other seven `_maybe_create_quantized_weight_buffers` call sites. Semantics preserved: the helper's internal `need_weight_transpose_cache` now equals the previously-intended `need_cache_colwise`.

## Notes
- Not caught earlier because MXFP4 runs only on MI355X (gfx950) while CI runs on MI300X (gfx942), so the FP4 path is never exercised in CI; and the mismatched kwarg is a runtime-only error that lint does not flag.


tests/unit_tests/backends/megatron/test_primus_turbo_quant_call_sites.py
```diff
@@ -0,0 +1,100 @@
+###############################################################################
```
RuibinCheung, line 1: Please remove this file.

RuibinCheung (changes requested):

WangLingxun:
Superseded by #882 — the disable_parameter_transpose_cache kwarg fix is already on main.

PR #864, opened 2026-07-08
feat: odc adapt

botaohu001:
This PR adapts the On-Demand Communication (ODC) framework to the AMD ROCm stack (MI300X) inside Primus. ODC reduces gradient-communication synchronization from per-microbatch to per-minibatch and overlaps single-sided push/pull communication with the backward pass, enabling LB-Mini load balancing (variable microbatch counts per rank) on top of a rocSHMEM-based point-to-point backend.

The rocSHMEM point-to-point ops are now provided by Primus-Turbo (primus_turbo.pytorch._C.odc_rocshmem_host / odc_rocshmem_gda) and consumed by _rocshmem_backend.py, instead of being built in-tree as ctypes .sos — ODC is therefore a pure-Python in-tree module now (the in-tree rocSHMEM sources and build_rocshmem_backend.sh are removed). This depends on Primus-Turbo PR #409. Validated to match the pure-Primus baseline on both single-node (host / XGMI IPC) and multi-node (GDA) runs.


.gitignore
Xiaoming-AMD, line 1: Why are these being ignored?

Xiaoming-AMD (changes requested):

Xiaoming-AMD (approved):

PR #869, opened 2026-07-12
feat(maxtext): support MaxText v26.4 with v26.3 backward compatibility

yeandy:
## Summary

Updates the Primus MaxText launcher to support **MaxText v26.4** while remaining backward-compatible with **v26.3 and earlier**. v26.4 relocated MaxText into a `maxtext.*` package (training loop at `maxtext.trainers.pre_train.train`, config at `maxtext.configs.pyconfig`, logging at `maxtext.utils.max_logging`), whereas older releases exposed `MaxText.*`. In addition, v26.4's `pyconfig` now hard-fails on unknown config fields instead of warning. This PR resolves the correct import layout at runtime so a single Primus checkout works against both v26.4 and v26.3 images.

## Changes

- **`maxtext_pretrain_trainer.py`**: add `_resolve_maxtext_train()` that imports `initialize`/`run` from `maxtext.trainers.pre_train.train`, falling back to legacy `MaxText.train`. The resolved module name is used as `argv[0]` for `initialize`.
- **Strip Primus-internal params before config export** (`file_sink_level`, `stderr_sink_level`, `sink_level`, `trainable`, `model`), since v26.4's `pyconfig` rejects unknown fields.
- **`patches/train_patches.py`**: add `_resolve_train_and_pyconfig()`; skip the override-forwarding patch gracefully if neither layout is importable.
- **`patches/logger_patches.py`**: add `_resolve_max_logging()` (`maxtext.utils.max_logging` → `MaxText.max_logging`).
- **Submodule**: bump `third_party/maxtext` to `release/v26.4` (`80f431d0`) to match the shipped v26.4 image.

## Testing

Validated end-to-end on **2 nodes (MI355X)**, both images reaching steady-state training with matching loss curves:

| Image | Result | Steady state |
|---|---|---|
| `rocm/jax-training:maxtext-v26.4-jax0.9.1-te2.12.0` | steps 0–4, exit 0 | ~1043 TFLOP/s/device |
| `rocm/jax-training:maxtext-v26.3` (backward compat) | steps 0–4, exit 0 | ~1066 TFLOP/s/device |

Loss descended identically (12.26 → ~12.10) on both, confirming the new `maxtext.*` and legacy `MaxText.*` code paths both work.

## Notes for reviewers

- The Python changes are non-breaking on their own (work with either image). The **submodule bump makes v26.4 the repo default**


yeandy:
@llying-001 @amd-fuyuajin Please review, thanks!

amd-fuyuajin:
@yeandy I had problem running this branch with `v26.4` image. I found the cause and fix. After fixing it, it runs with both v26.4 and v26.3 images. No performance difference observed.

## Root Cause
Not the image, not RCCL, not the branch's dual-API code. It was a **config validation error unique to MaxText v26.4**.

- v26.4 introduced a strict `pydantic` config model (`MaxTextConfig`) with this rule (`third_party/maxtext/src/maxtext/configs/types.py:2929`):
  ```
  if hardware == "gpu" and packing and attention == "cudnn_flash_te" and max_segments_per_seq <= 0:
      raise ValueError("max_segments_per_seq must be set when using TransformerEngine attention and packing")
  ```
- Primus defaults satisfy all four conditions: `hardware: gpu`, `packing: true`, `attention: cudnn_flash_te` (AMD default), and `max_segments_per_seq` defaulting to `-1`.
- **v26.3 had no such check**, so the identical config passed — hence v26.3 worked and v26.4 didn't.

The actual `pydantic.ValidationError` was raised inside `pyconfig.initialize()`, caught by Primus and re-raised, but the message was written **only to the loguru file sink** (`output/.../logs/pre_trainer/rank-0/debug.log`), never to stdout — which is why the console showed a bare `exit 1`.

## Fix
Added one line to the shared MaxText base config `primus/configs/modules/maxtext/pre_trainer.yaml` (next to `packing: true`):

```yaml
max_segments_per_seq: 32
```

`32` matches every upstream MaxText GPU model config (e.g. `configs/gpu/models/llama3_8b.yml`). It's a long-standing MaxText param, so it's harmless on v26.3 and keeps dual-API compatibility. Placing it in the base config covers all MaxText configs (BF16 and FP8 both inherit it; the rule is quantization-independent).

wenxie-amd (approved):

PR #933, opened 2026-07-29
fix: bump PRIMUS_TURBO_COMMIT to 69deeab7 to fix loss NaN

zhenhuang12:


WangLingxun:
flydsl 0.3.0 (released 2026-08-01) removed `flydsl/expr/buffer_ops.py` and
`flydsl/expr/vector.py`, which aiter, primus-turbo and Primus's own
DeepSeek-V4 kernels all import at module scope.

primus-turbo declared `flydsl>=0.2.0` with no upper bound, so the turbo install
(a separate pip invocation from aiter's) resolved to 0.3.0. `69deeab7`
(2026-07-29) still carries that unbounded requirement, so this bump would
rebuild the image and land on 0.3.0 again.

Upstream already fixed it in
[Primus-Turbo 2f622de6](https://github.com/AMD-AGI/Primus-Turbo/commit/2f622de6a29ab711925e8fe6bcc763be76ac2699)

PR #938, opened 2026-07-30
fix(megatron): make device_id init_process_group patch opt-in (AIMA-227)

gphuang:
## Summary

Makes the `device_id` injection into `torch.distributed.init_process_group`
opt-in via `enable_init_process_group_device_id` (default `false`), instead of
applying to every FSDP2 run.

Tracking: AIMA-227. The patch was introduced by
[PR #856](https://github.com/AMD-AGI/Primus/pull/856) and was suspected of
costing llama3-70B throughput. It does not cost throughput — it costs memory.

### Measured cost of the patch

`llama3_70B-BF16-pretrain.yaml`, mock data, FSDP2, TP/PP/EP=1, mbs 3 / gbs 24,
seq 8192, 1 node x 8 MI355X, ROCm 26.5. Arms interleaved ON/OFF/ON/OFF on one
node so node/thermal drift cannot masquerade as a patch effect; 50 iterations
each, 2 repeats per arm.

| Metric | patch ON vs OFF |
|---|---|
| Steady-state iteration time | **−0.07 %** (ON marginally faster; within run-to-run noise) |
| Throughput (TFLOP/s/GPU) | within noise, same direction |
| Megatron init | **+9.9 s** |
| Peak VRAM (`rocm-smi`) | **+16.03 GiB/GPU** |
| Torch max reserved (torch allocator) | **0 MB** (byte-identical) |

Both arms: 50/50 iterations, 0 NaN, 0 skipped, loss converging identically.

The extra 16 GiB is **invisible to the torch caching allocator** (max reserved
is byte-identical across arms), so it is not activations or parameters — it is
communicator-side memory from eagerly building the world PG plus every Megatron
sub-group. The old docstring estimated ~768 MiB; the measured cost is ~21x that,
so the docstring is corrected here too.

### Why opt-in rather than removed

With the patch disabled, the deadlock it guards against did not reproduce:

- ROCm 26.5, patch off: 50/50 iterations, twice, no hang.
- Stock `rocm/primus:v26.3` (bundled Primus `45955d45`, which has no
  `distributed_init_patches.py` at all): 20/20 iterations, `rc=0`, no hang,
  loss converging normally. First-iteration loss agrees with the 26.5 arms to
  five significant figures, so both are running the same model and data.

Without the patch PyTorch simply emits the warning the patch exists to silence
(`barrier(): using the device under current context. You can specify device_id
in init_process_group ...`) and proceeds.

Since version was **not** the discriminating variable, a ROCm-version gate
would be wrong. But the evidence is single-node with TP/PP/EP=1, so the patch is
kept and gated rather than deleted.

### Scope not covered by this evidence

- **1 node x 8 GPUs only** — the classic RCCL device-mapping deadlock is a
  multi-node phenomenon; the allocation QoS capped at one node.
- **TP = PP = EP = 1** — this recipe is pure FSDP2 data parallelism, so most
  Megatron sub-groups are degenerate. A TP/PP/EP-heavy recipe builds real
  sub-group communicators and would likely pay a larger eager-init cost.
- **One recipe** (`llama3_70B-BF16`).

### Behaviour change

60 recipe YAMLs set `use_torch_fsdp2: true` and therefore stop applying the
patch by default. MI300X/MI325X recipes were paying the 16 GiB for nothing —
the existing ODC comment in this file already states the MI355X deadlock "does
not trigger" on MI300X. The MI355X FSDP2 recipes are the ones whose behaviour
genuinely changes.

**Open question blocking merge (hence draft):** should any shipped MI355X recipe
set `enable_init_process_group_device_id: true`? That needs the original repro
conditions — node count, recipe, parallelism — from the patch author
(@luiza-amd). If the deadlock was multi-node, the multi-node recipes should set
it true and this PR should carry that change.

## Test plan

- [x] 6 new unit tests in
      `tests/unit_tests/backends/megatron/patches/test_distributed_init_patches.py`:
      full `condition` truth table (absent flag, FSDP2-only, explicit opt-in,
      DDP rejection, ODC carve-out) plus an assertion that the shipped default is
      `false`, so the default cannot silently flip back.
- [x] Test is ordering-independent: sibling suites clear the process-wide
      `PatchRegistry`, so the fixture reloads the patch module when needed.
      Verified passing both alone and after `tests/unit_tests/core/patches`.
- [x] `pytest tests/unit_tests/core/patches tests/unit_tests/backends/megatron/patches`
      → 4 failed, 68 passed. The 4 failures are **pre-existing** in
      `tests/unit_tests/core/patches/test_patch.py`, reproduced on a pristine
      `origin/main` worktree at `a89d28b5` (4 failed, 62 passed) before this change.
- [x] `black==24.8.0` (line-length 110), `isort==5.13.2 --profile black`,
      `autoflake==2.3.1` all clean on the changed files.
- [x] End-to-end: patch-off arms confirmed via the patch's own banner being
      absent from the training log, and patch-on arms via it being present.
- [ ] Confirm with the patch author whether a multi-node MI355X recipe needs the
      flag set true.

### DIFF
--- a/primus/backends/megatron/patches/distributed_init_patches.py
+++ b/primus/backends/megatron/patches/distributed_init_patches.py
@@ -5,17 +5,30 @@
 ###############################################################################
 
 """
-Distributed initialization patches (FSDP2 only).
+Distributed initialization patches (opt-in, FSDP2 only).
 
 Patches Megatron's _initialize_distributed to pass device_id to
 torch.distributed.init_process_group. Without device_id, RCCL guesses
-the GPU-to-rank mapping, which causes deadlocks on MI355X after the
-first FSDP2 iteration.
+the GPU-to-rank mapping, which was reported to cause deadlocks on MI355X
+after the first FSDP2 iteration.
 
-This patch is gated on use_torch_fsdp2 because device_id triggers eager
-RCCL communicator creation for the world PG and all ~26 Megatron sub-groups,
-consuming ~768 MiB of additional GPU memory. Under DDP this extra allocation
-pushes large models over the GPU memory limit.
+device_id triggers eager RCCL communicator creation for the world PG and all
+~26 Megatron sub-groups. That is expensive, so the patch is opt-in via
+enable_init_process_group_device_id (default false):
+
+- +16.0 GiB/GPU peak VRAM and +9.9 s startup, measured on llama3-70B BF16
+  FSDP2, 1 node x 8 MI355X, TP/PP/EP=1, ROCm 26.5. The memory is invisible to
+  the torch caching allocator (max reserved is byte-identical across arms), so
+  it is communicator-side, not activations or parameters.
+- No throughput effect there (-0.07 %, within run-to-run noise over two
+  interleaved repeats per arm).
+- Under ODC the eager communicators also serialize XGMI copy streams; see the
+  condition below.
+
+The deadlock this guards against did not reproduce on that recipe with the
+patch disabled, on either ROCm 26.5 or 26.3, so it must not be assumed
+necessary everywhere. It has not been tested multi-node or with TP/PP/EP > 1,
+which is why the patch is kept rather than removed. See AIMA-227.
 """
 
 import os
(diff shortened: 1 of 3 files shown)


gphuang:
Superseded by a freshly opened PR for the same change; closing to keep one active review thread.

PR #807, opened 2026-06-28
feat(flux): core runtime + Megatron adapter scaffolding

jasainio:
Base of an 18-PR series splitting the Flux diffusion-training feature (training Flux, a DiT text-to-image diffusion model, on Primus/Megatron) out of one large branch for reviewability. Targets `feat/flux/ci-env` and auto-retargets to `main` once that merges. Every other content PR stacks on this one.

## What this changes
The shared runtime scaffolding the rest of the feature builds on: core runtime state + train-runtime wiring, the Megatron adapter and base/pretrain trainers, and a patch auto-loader (`patches/__init__.py`) that imports every `*_patches.py` in the package on import, so each later layer only drops in its own patch file with no registry edit. Also carries the shared test root (`tests/conftest.py`, `tests/utils.py`) and a one-line `.gitignore` change (the bare `data` ignore → root-anchored `/data/*`, and nothing else) that stops Git from ignoring the in-repo `data/` source and config directories the later layers add. Kept deliberately small so it can land first.

## Dependencies
Sequenced after the CI-pins PR (`feat/flux/ci-env`); no functional dependency on the turbo bump.

## Test plan
`pytest tests/unit_tests/core tests/unit_tests/backends/megatron`; lint/pre-commit clean. Validated locally on an AMD GPU container: 58 passed.

## Files
39 (core runtime, Megatron adapter/trainers, base patch loader, shared test root, `.gitignore`).

PR #939, opened 2026-07-30
fix(megatron): gate device_id process-group init (AIMA-227)

gphuang:
## Summary

Tracking: AIMA-227. Introduced through [PR #807](https://github.com/AMD-AGI/Primus/pull/807) and merged to `main` through [PR #856](https://github.com/AMD-AGI/Primus/pull/856).

Makes the `device_id` injection into `torch.distributed.init_process_group` opt-in through `enable_init_process_group_device_id` (default `false`).

Passing `device_id` forces eager RCCL communicator creation and adds about 16 GiB/GPU outside the torch caching allocator. This exhausts the remaining VRAM on Llama-70B FP8 and causes the AIMA-227 throughput regression.

An exact same-HEAD FP8 check recovered the 16 GiB, matched the old memory baseline, and completed 10/10 iterations with the patch disabled. Llama also completed without the patch on stock 26.3 and 26.5.

The patch is gated instead of deleted because the historical failure involved Flux-schnell on Megatron with TransformerEngine. If that workload still deadlocks on v26.5 with the patch disabled, the affected Flux recipes should opt in explicitly.

## Test plan

- [x] 6 unit tests cover the gating truth table and default value.
- [x] Llama BF16 patch ON/OFF: valid loss, 50/50 iterations per arm.
- [x] Llama FP8 patch OFF: old memory baseline restored, 10/10 iterations.
- [x] Stock 26.3 without the patch: 20/20 iterations.
- [x] Flux-schnell Megatron/TE on v26.5: patch ON versus OFF.


gphuang:
## Note: original repro question folded into PR summary

The original repro question has been folded into the [PR summary](https://github.com/AMD-AGI/Primus/pull/939); the current request is tracked in the [historical Flux-schnell question](https://github.com/AMD-AGI/Primus/pull/939#issuecomment-5131533397).

gphuang:
## Note: Llama result folded into PR summary

The detailed Llama patch-OFF result has been folded into the [PR summary](https://github.com/AMD-AGI/Primus/pull/939); it does not cover the historical Flux-schnell workload tracked in the [current question](https://github.com/AMD-AGI/Primus/pull/939#issuecomment-5131533397).

gphuang:
## Question: historical Flux-schnell repro

@amital-amd — could you share the exact image/commit, config, node count, parallelism, and hang point from the original Flux-schnell Megatron/TE deadlock? We need one v26.5 patch ON/OFF liveness check before merging.

amital-amd:
@gphuang I did not have those details. Could not find them from logs, probably deleted them during cleanup of dir to save space.

eshaw2:
# Note: No hangs observed for Flux model on v26.5 with patch off

- Test done: Flux-schnell Megatron/TE on v26.5: patch ON versus OFF

Swept through the full set of Flux FSDP configs with patch on and off on v26.5 and v26.5rc_0721. No hangs were observed over 6 configs (x2 iterations, x2 images). Also swept Llama as additional validation.

Patch is likely safe to remove, but safe to keep as opt-in.

eshaw2:
# fix(megatron): remove the device_id init_process_group patch

The patch cost +16.0 GiB/GPU peak VRAM and +9.9 s of startup on llama3-70B
BF16 FSDP2 (1 node x 8 MI355X, TP/PP/EP=1, ROCm 26.5) for no throughput
change (-0.07 %, within noise). The memory is invisible to the torch caching
allocator -- max reserved is byte-identical across arms -- so it is
communicator-side, from eagerly creating the world PG and every Megatron
sub-group communicator.

The deadlock it guarded never reproduced with the patch disabled on Primus
26.5. A config sweep with the patch on completed 20/20 iterations
at 98.82 % occupancy, and patch-off arms pass at 97.5 %.

## Behavior changes 
On main today the patch applies whenever a recipe sets `use_torch_fsdp2: true` without `enable_odc: true`. That is roughly 60 shipped example recipes under examples/megatron/configs/{MI300X,MI325X,MI355X} (the two ODC recipes there were already carved out). After removal, `torch.distributed.init_process_group` is called exactly as upstream Megatron calls it, with no `device_id,` on every recipe -- `device_id` appears nowhere else in primus/. The ODC carve-out and the FSDP2 gate become moot and vanish with the condition.

wenxie-amd (approved):

PR #986, opened 2026-08-13
feat(megatron): select MXFP4 forward precision

gphuang:
## Summary

Tracks [Issue 220](https://github.com/AMD-AGI/tiger-training-internal/issues/220).

Add independent MXFP4 forward precision selection so late recovery runs can use FP8 or BF16 forward GEMMs while keeping MXFP4 data- and weight-gradient GEMMs.

- Add fail-closed `mxfp4_forward_precision` config plumbing and runtime mode census.
- Reuse existing FP8 and BF16 operators; no new kernel implementation.
- Cover direct autograd, compile, Flux integration, config validation, and audit modes.

## Test plan

- [x] Python compile check
- [x] `git diff --check`
- [ ] Targeted CPU-safe unit tests in the v26.5 container
- [ ] MI355X direct linear and Flux 535M forward/backward tests
- [ ] Option 6/7 resume smoke with exact mode census

PR #987, opened 2026-08-13
test(megatron): cover compiled MXFP4 backward

gphuang:
## Summary

Closes the post-merge review gap from PR #986.

Adds regression coverage for FP8/BF16 forward with MXFP4 backward through `torch.compile`, including exact eager/compiled output and gradient comparisons. It also covers YAML-to-`FluxConfig` precision propagation and runtime mode-census emission.

## Test plan

- [x] 35 config and audit tests passed
- [x] Compiled-backward checks passed for FP8 and BF16 forward modes
- [x] Flux 535M forward-only integration test passed
- [x] `black` formatting verified
- [x] GPU memory returned to idle

### DIFF
--- a/tests/unit_tests/backends/megatron/diffusion/training/test_flux_model_creation.py
+++ b/tests/unit_tests/backends/megatron/diffusion/training/test_flux_model_creation.py
@@ -245,6 +245,23 @@ class TestFluxModelCreation:
         assert config.fp8_dot_product_attention is False
         assert config.fp8_multi_head_attention is False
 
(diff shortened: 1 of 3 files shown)


gphuang:
## Verdict: compiled-backward coverage is ready to merge

No blocking findings in the full base-to-head diff.

- Exact eager/compiled outputs and input/weight gradients passed for FP8 and BF16 forward with MXFP4 backward.
- Config propagation, runtime audit emission, and the Flux integration path passed on GPU.
- Required CI is green; GPU CI is skipped, with the equivalent targeted GPU gate run on the allocated node.

Two optional hardening ideas—testing a future non-default backward mode and pinning preshuffle state inside the test—do not affect the current production path.

PR #994, opened 2026-08-18
feat(deepseek-v4): CP adaption and gfx942 (MI308X/CDNA3) adaptation — 128k CP + full-model multi-node SFT 

aaab8b:
## What                                                                                                     
                                                                                                              
  Make DeepSeek-V4-Flash trainable on **gfx942 / CDNA3** (MI308X). The upstream                               
  sparse-MLA kernels target gfx950/CDNA4; this PR adapts them plus the long-context                           
  and multi-node paths so the full model trains on CDNA3.                                                     
                                                                                                              
  ## gfx942 kernel fixes                                                                                      
  - **DSA sparse-MLA backward LDS overflow** (compile-time): the 512-latent Q tile                    
    plus Triton's stage double-buffering asks for 73728 B > the 64 KB CDNA3 LDS         
    budget. Fixed with `num_stages=1` (`PRIMUS_DSA_BWD_NUM_STAGES`), an                        
    `amd_pingpong_disabled()` compile scope that disables primus_turbo's global                               
    ping-pong / async-copy knobs (which double-buffer LDS and are a measured                             
    pessimization for these kernels), and tunable tile sizes as a fallback.                                   
  - Forward pinned to the non-ping-pong schedule; backward skips provably-zero
    rope MMAs.                                                                                                
                                                                                                              
  ## Long-context / context parallelism                                                                       
  - CSA/HCA CP: left-boundary KV exchange (P2P) + global compressed-pool                                      
    all_gather (`deepseek_v4_cp.py`).                                                                         
  - Indexer scoring gains `Q_OFFSET` so visibility is judged on the query's                                   
    **global** position under CP (0 without CP → identical to before).                                        
  - Streaming chunked top-K (`PRIMUS_INDEXER_TOPK_CHUNK`) avoids materializing the                            
    full `[B,S,P]` score row; full-Triton indexer path avoids an eager einsum that                            
    would materialize `[B,S,H,P]`.                                                                         
                                                                                                              
  ## Multi-node full-model SFT                                                                                
  - `NNODES=2/3` branches with PP layouts; optimizer CPU offload with a tunable                               
    fraction; socket-network note: `NCCL_ALGO=Ring` fixes a Tree-over-TCP                                     
    cross-node all_reduce deadlock on fabrics without RDMA.          
                                                                                                              
  ## Scripts & docs                                                                                           
  One-click launchers `examples/deepseek-v4/gfx942/run_128k_dense_hca_csa.sh` and                             
  `run_full_4k_multinode.sh`, plus READMEs.                                                                   
                                                                                                              
  ## Verification                                       
  - 128k, 4-layer (CP=8/TP=1): 42.30 GB peak.
  -  Full model, 4k, 3 nodes: 10-step and 100-step runs, loss 11.90 → 9.19
    monotone, grad norm 20 → 2.8, **0 nan iterations**, exit 0.
                                                        
  All changes are **gfx942-gated or env-var-defaulted-off** and do not change
  gfx950 numerics or behavior.
                                                        
  ## Test plan                                          
  - [x] `pre-commit run --all-files` green on the changed files
  - [x] 128k 4-layer smoke on 1×8 MI308X                
  - [x] full-model 4k 10-step & 100-step on 3×8 MI308X (no NaN)

### DIFF
--- a/examples/deepseek-v4/gfx942/README.md
+++ b/examples/deepseek-v4/gfx942/README.md
@@ -1,0 +1,260 @@
+# DeepSeek-V4 SFT on gfx942 (MI308X / CDNA3)
+
+Runs a 4-layer DeepSeek-V4 SFT at **128k context** exercising **all three** V4 attention
+branches — dense+SWA, CSA, HCA — on a single 8×MI308X node.
+
+Upstream's V4 support targets gfx950 / CDNA4 and, as shipped, only ever ran at
+`seq_length=4096`. Several things that are invisible at 4k become hard blockers at 128k.
+This directory carries the fixes and a one-click launcher.
+
+```bash
+# inside a rocm/primus:v26.5-pytorch2.12-te2.15 container, from the repo root
+bash examples/deepseek-v4/gfx942/run_128k_dense_hca_csa.sh
+```
+
+The script is self-contained: it copies Megatron-LM out of the image (or fetches the
+submodule), builds the SFT dataset on first run, and resolves every path relative to
+itself. The only external requirement is a **local DeepSeek-V4 tokenizer directory** —
+point `V4_TOKENIZER` at it, or leave it at `/apps/DeepSeek-V4-Flash`. Only
+`tokenizer.json` / `tokenizer_config.json` are read.
+
+## Scope — read this before quoting any number
+
+* **Weights are randomly initialised.** No DeepSeek-V4 Megatron checkpoint exists, and
+  Megatron-Bridge registers importers only for V2-Lite/V2/V3. This is *SFT-shaped training
+  from scratch*, not fine-tuning. The tokenizer and the prompt/response loss mask are real.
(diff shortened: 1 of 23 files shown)


aaab8b:
Superseded by #1047.

This PR was opened from a fork, so its CI could not access repository secrets and the
docker build step could not run. #1047 carries the same work from an in-repo branch
(`zhaobing/dev/deepseek-v4-gfx942-cp`), rebased onto current `main` — which had advanced
59 commits and touched the same V4 files, so three files needed conflict resolution.

Two things there are worth a look even if you already reviewed this PR:

* the indexer key pool now gathers before rotating, because `_rotate_keys` treats a key's
  tensor index as its global compressed position — rotating the local pool first is
  correct on rank 0 and the wrong phase on every other rank, silently;
* a fix for `_AllGatherPool.backward` writing into autograd's `grad_output`, plus four
  bitwise (`torch.equal`) test groups over the CP collectives and metadata that found it.

Closing in favour of #1047; review history stays here.

PR #1054, opened 2026-08-27
[JAX] fix XLA_FLAGS precedence inversion, add XLA_FLAGS_APPEND, document flag layering

yeandy:
## Problem

`apply_env_defaults()` runs after the config's `env:` block, and the old per-flag merge always favored the managed value — so a config setting `XLA_FLAGS: "--xla_gpu_autotune_level=5"` silently got an effective `=4`. Precedence was inverted for `XLA_FLAGS`.

`setdefault` cannot fix it: the JAX images bake `XLA_FLAGS`, so the key is always present and `setdefault` no-ops, dropping all nine managed flags — including the autotune level that prevents NaN loss on fp8 MoE runs.

## What changed

- Replace the per-flag `OrderedDict` merge with a plain append. XLA honors the last occurrence, so this is simpler and also fixes values containing spaces (`--xla_gpu_enable_command_buffer=''`).
- `mark_config_owned()` tracks keys from the config's `env:` block; managed append-mode defaults now step aside for them and log that they did.
- Add `XLA_FLAGS_APPEND` to override one flag without taking ownership of the whole string. Applied in `BackendAdapter.apply_env_defaults()` so it also reaches backends declaring no defaults (MaxDiffusion).
- Docs: new `docs/02-user-guide/environment-and-xla-flags.md`; fixed the JAX/XLA reference section (still credited the removed prepare hook, had no `XLA_FLAGS` row); documented the top-level `env:` block.

Precedence, highest last: inherited → backend defaults → per-config `env:` → `XLA_FLAGS_APPEND`.

## Verification

18 new tests in `test_env_registry.py`, including a named regression test for the inversion; `453 passed, 1 skipped` across the core and maxtext suites.

Also ran the real path (`_apply_config_env` + `MaxTextAdapter().apply_env_defaults()`) with a baked `autotune_level=0`: append gives an effective `=5` with managed and unrelated baked flags intact, a config-set `XLA_FLAGS` skips the defaults and logs it, and `DUMP_HLO=1` yields `--xla_dump_to`. In a JAX image (CPU-only), the managed and duplicate-flag strings both parse while a bogus flag dies with `Unknown flag in XLA_FLAGS`, and `--xla_dump_to=/tmp/xa --xla_dump_to=/tmp/xb` wrote 0 and 30 files, reconfirming last-occurrence-wins.

## Blast radius

Only the six MaxDiffusion examples set `XLA_FLAGS`, and that backend declares no managed defaults, so they are unaffected in practice. No MaxText config sets it. Anyone relying on managed defaults beating a config's own `XLA_FLAGS` now gets the reverse — that is the fix, and it is logged.

## Follow-ups

- `Dockerfile.jax-v26.4/.5` still bake `--xla_gpu_autotune_level=0`: latent under append, but contradicts the docs.
- Multi-node run to confirm log ordering under `primus-cli`, and that the `first_collective_call` timeouts change behavior (only parsing is confirmed).


Xiaoming-AMD (approved):

i-chaochen (commented):
https://docs.nvidia.com/jax-toolbox/performance-profiling/gpu-performance#effect-details-o1

you can turn the `O1`  ON or specify these FLAGs


```
--xla_gpu_enable_latency_hiding_scheduler=true
--xla_gpu_enable_pipelined_all_reduce=true
--xla_gpu_enable_pipelined_all_gather=true
--xla_gpu_enable_pipelined_reduce_scatter=true
```

PR #1067, opened 2026-09-01
fix(runner): honor PRIMUS_SKIP_PIP in the Megatron pretrain hook

gphuang:
## Summary

The Megatron **pretrain** install hook has no `PRIMUS_SKIP_PIP` guard, unlike its posttrain Megatron and MaxDiffusion siblings, so there is no way to stop it on an image that already ships the stack. This adds the same guard, using the same three-way check (`unset` / `0` / `false` all mean "do not skip").

On a ROCm image the missing guard is fatal rather than merely wasteful. `requirements-megatron.txt` pins the Qwen3.5 gated-delta-net dependencies, and `causal-conv1d~=1.5` resolves to a version with no wheel. pip therefore builds it from source, and the isolated build environment resolves a fresh `torch` against the image's pinned ROCm build:

```
ERROR: Cannot install torch because these package versions have conflicting dependencies.
    The user requested torch
    The user requested (constraint) torch==2.12.0+rocm7.15.0a20260720
ERROR: ResolutionImpossible
```

The hook exits non-zero, `prepare_experiment.sh` fails, and the run dies before training starts — over two packages the model under test never imports. This currently breaks any Megatron pretrain launched through `primus-cli direct` on a ROCm image, which includes the MLPerf Flux launcher at `examples/mlperf/flux1/megatron/run_and_time.sh`.

## Test plan

- [x] `bash -n` clean
- [x] `PRIMUS_SKIP_PIP=1` skips and exits 0
- [x] `PRIMUS_SKIP_PIP=0` and unset both still install (guard does not change the default path)
- [x] Reproduced the original failure on `rocm/primus:v26.5` (MI355X, Flux 12B MXFP6 MLPerf recipe)
- [ ] End-to-end rerun of that recipe with the guard set — waiting on a node

### DIFF
--- a/runner/helpers/hooks/train/pretrain/megatron/00_install_requirements.sh
+++ b/runner/helpers/hooks/train/pretrain/megatron/00_install_requirements.sh
@@ -13,10 +13,10 @@ PRIMUS_ROOT="${PRIMUS_PATH:-${DEFAULT_PRIMUS_ROOT}}"
 # Standard pip-skip switch, matching the posttrain Megatron and MaxDiffusion
 # hooks. An image that already ships the stack has nothing to gain here, and on
 # a ROCm image it can actively lose: these requirements are the Qwen3.5
-# gated-delta-net deps, and causal-conv1d ships no wheel, so pip builds it from
-# source and the build environment resolves a fresh `torch` against the image's
-# pinned ROCm build. That conflict is unresolvable and aborts the run before
-# training starts, for packages the model under test never imports.
+# gated-delta-net deps, the version causal-conv1d~=1.5 resolves to ships no
+# wheel, and its isolated source build resolves a fresh torch against the
+# image's pinned ROCm build. That conflict is unresolvable, so the run dies
+# before training starts over packages the model under test never imports.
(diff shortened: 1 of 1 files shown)


gphuang:
Closing — not needed after all, and the reasoning behind it was wrong.

I opened this because the Megatron pretrain hook aborted an issue 306 run before training started. I attributed that to `causal-conv1d~=1.5` shipping no wheel and its source build fighting the image's pinned ROCm torch. Both are wrong. The ROCm image already ships a satisfying `causal_conv1d 1.5.0.post8`, and the hook's own log reports it as already satisfied; the failure never reaches a compiler.

What actually happened is on the caller's side. Our dev container exports a `PIP_CONSTRAINT` file pinning torch to its local version label (`2.12.0+rocm7.15.0a20260720`), which no index carries. `~=1.5` also admits 1.7.0, so pip can reach past the installed build to the newer sdist, and the isolated build environment then has to resolve `torch` from an index and cannot. Pinning `causal-conv1d` to the shipped version in our own constraint file removes the build path entirely, with no change to Primus.

Proved by A/B varying only the constraint file: strict gives `ResolutionImpossible`, loose resolves build dependencies. So the hook needs no escape hatch for this case. If someone wants a skip switch for its own sake, that is a separate argument and this PR's justification should not be used for it.

PR #1069, opened 2026-09-01
fix(flux): wait on the warmup's gradient reduce-scatter

gphuang:
## Summary

Every Flux run with `mlperf_mode: true` and `warmup_train_steps > 0` dies during the first real training step, before a single iteration completes. This makes the warmup wait on the gradient reduce-scatter it dispatches.

The MLPerf warmup runs at the pre-data boundary, which is before Megatron's `train()` executes. `train()` is where `config.finalize_model_grads_func` gets assigned, and that callback is the only caller of `finish_grad_sync()`. So with `overlap_grad_reduce: true` a warmup backward dispatches the data-parallel reduce-scatter and nothing ever waits on it. The handle is still outstanding when the first real step dispatches its own, and Megatron asserts:

```
File ".../megatron/core/distributed/param_and_grad_buffer.py", line 530, in start_grad_sync
    self.grad_reduce_handle is None
AssertionError: Should not have multiple communication calls outstanding at once
```

### The fix

Install `finalize_model_grads` for the duration of the warmup loop and restore whatever was there before, so the boundary warmup does not change what `train()` later sees. The `None` guard leaves the in-`train_step` warmup path used by development recipes untouched, since by then `train()` has already installed the callback — which is why only MLPerf mode was affected.

### Why the escape hatch does not cover this

`start_grad_sync` no-ops when a handle is outstanding *and* `is_first_batch` is still set. Warmup step 1 populates the ready counts without dispatching; the `reset()` at the head of warmup step 2 records the golden counts and clears `is_first_batch`; warmup step 2 then dispatches for real. By the first training step the flag is long gone, so the assertion fires.

## Test plan

- [x] Two unit tests in `tests/unit_tests/backends/megatron/test_mlperf_patches.py`: the warmup steps see a grad-finalize callback, and a callback the caller already set is preserved rather than replaced. Both fail without the production change.
- [x] Full `test_mlperf_patches.py` passes (46 tests) on an MI355X dev container.
- [x] Bisected on 8x MI355X with the Flux 12B MXFP6 MLPerf recipe. The FP8 MLPerf recipe fails identically, so this is not precision-specific; `mlperf_mode: false` passes; `warmup_train_steps: 0` passes.
- [x] With this change the full MLPerf recipe at `warmup_train_steps: 2` trains 20 iterations clean, and a full convergence run is in flight.

### DIFF
--- a/primus/backends/megatron/patches/mlperf_warmup_patches.py
+++ b/primus/backends/megatron/patches/mlperf_warmup_patches.py
@@ -270,19 +270,37 @@ def _run_warmup_and_restore(
     # ---- 3b. Save LR scheduler state (NeMo never steps the scheduler during warmup) ----
     saved_lr_num_steps = opt_param_scheduler.num_steps
 
+    # ---- 3c. Install Megatron's grad-finalize callback for the warmup steps ----
+    # At the pre-data boundary `train()` has not run yet, so it has not assigned
+    # config.finalize_model_grads_func. That callback is the only caller of
+    # finish_grad_sync(), so without it a warmup backward dispatches the
+    # data-parallel reduce-scatter and nothing ever waits on it. The handle is
+    # still outstanding when the first real step dispatches its own, and
+    # Megatron asserts "Should not have multiple communication calls
+    # outstanding at once" before a single iteration completes.
+    saved_finalize = config.finalize_model_grads_func
+    if saved_finalize is None:
+        from megatron.core.distributed import finalize_model_grads
+
+        config.finalize_model_grads_func = finalize_model_grads
+        _log("Installed finalize_model_grads_func for warmup (unset at this boundary)")
+
     # ---- 4. Run warmup steps with synthetic data ----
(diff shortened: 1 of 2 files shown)


gphuang:
## Note: author review — ready to merge

Reviewed the full base-to-head diff against `fix/flux-eval-correctness`. No blocking findings.

One non-blocking fix-up, now pushed: the second test declared its nested helper with no blank line before it, which `black` would have reformatted.

Checks run in an MI355X dev container on the exact head commit: `pytest tests/unit_tests/backends/megatron/test_mlperf_patches.py` passes 46 tests, `black --check` clean, `isort --profile black --check-only` clean, `autoflake --check` clean. The repository's pre-commit set has no flake8 hook, so that was not run.

Both new tests were verified to fail with the production change reverted, so they are not vacuous. The end-to-end proof is the 20-iteration MXFP6 MLPerf run described in the PR body, which is the same recipe that previously died before iteration 1.

PR #1081, opened 2026-09-03
fix: exclude FP8 quantized weights from MXFP4 de-osc

hann-wang:
## What's wrong

MXFP4 weight-deosc process selects eligible weights by looking for `quantized_weight_buffer`. FP4 and FP8 quantized modules both have this buffer when gradient accumulation enabled.

## What's changed

* MXFP4 quantized module creates an empty buffer in `float4_e2m1fn_x2` dtype when gradient accumulation disabled.
* Check the dtype of `quantized_weight_buffer` in MXFP4 weight de-osc.

@HuangWei-95


Xiaoming-AMD:
Please resolve the conflicts.

PR #1083, opened 2026-09-03
testintel: add missing test for AvgDown3D.forward (created by linxwang)

jiagaoxiang:

No existing test in tests/unit_tests/backends/diffusion (or anywhere else in the repo) instantiates AvgDown3D or imports vae2_2.py at all; the pad->view->permute->view->view->mean sequence that implements grouped spatiotemporal average-downsampling, including the temporal padding branch (pad_t>0) and the group_size averaging when in_channels*factor != out_channels*group_size trivially, is completely unexercised.

This PR adds a focused unit test for `AvgDown3D.forward`; it does not change production code.

### DIFF
--- a/tests/unit_tests/backends/diffusion/test_wan_vae_avgdown3d.py
+++ b/tests/unit_tests/backends/diffusion/test_wan_vae_avgdown3d.py
@@ -1,0 +1,54 @@
+###############################################################################
+# Copyright (c) 2025, Advanced Micro Devices, Inc.
+#
+# See LICENSE for license information.
+###############################################################################
+
+import pytest
+import torch
+
+from primus.backends.diffusion.models.wan.vae2_2 import AvgDown3D
+
+
+def test_forward_averages_grouped_channels_without_padding():
+    # in_channels=1, factor=2*1*1=2 -> out_channels=1 needs group_size=2
+    module = AvgDown3D(in_channels=1, out_channels=1, factor_t=2, factor_s=1)
+
+    x = torch.arange(4, dtype=torch.float32).view(1, 1, 4, 1, 1)
+    out = module(x)
+
+    assert out.shape == (1, 1, 2, 1, 1)
+    # consecutive pairs of frames along t are averaged together
+    expected = x.view(1, 1, 2, 2, 1, 1).mean(dim=3)
+    torch.testing.assert_close(out, expected)
+
+
+def test_forward_pads_time_dimension_to_multiple_of_factor_t():
+    module = AvgDown3D(in_channels=1, out_channels=1, factor_t=2, factor_s=1)
+
+    # t=3 is not a multiple of factor_t=2, so one zero frame is left-padded.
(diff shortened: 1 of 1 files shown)


WangLingxun:
Please revise the spatial/channel test before marking this PR ready. `test_forward_downsamples_spatial_and_channel_dims` currently asserts only the output shape, so it would not catch an incorrect `permute` order or `group_size` channel mapping. Use a deterministic position-distinguishable input (for example, `torch.arange`) and compare the complete output against an independent numerical reference for `factor_s=2` with `in_channels != out_channels`. The reference should verify both spatial block grouping and the contiguous channel/group mapping, rather than reusing `AvgDown3D` internals. Keep the existing temporal front-padding and non-divisible-constructor coverage. This is regression coverage; no current production bug is asserted.

jiagaoxiang:
**TestIntel PR Steward:** Addressed @WangLingxun's review comment in `4d37455`.

`test_forward_downsamples_spatial_and_channel_dims` now uses `factor_t=1, factor_s=2, in_channels=2, out_channels=4` with a `torch.arange` (position-distinguishable) input, and asserts the full output against an independent, hand-computed reference (no calls into `AvgDown3D`'s own pad/view/permute/view/view/mean sequence). The reference derives, for every input element, which `(output_channel, spatial_block)` slot it lands in from `factor_t`/`factor_s`/`group_size` alone, accumulates the sums, and checks each group has exactly `group_size` contributors before comparing with `torch.testing.assert_close`. This exercises both the spatial block grouping (h/w downsample) and the contiguous channel/group mapping, and would fail on an incorrect `permute` order or `group_size` mapping.

The existing temporal front-padding and non-divisible-constructor tests are unchanged, as requested.

I verified the reference formula against the actual module logic with a standalone reshape/permute simulation before pushing, so this should be correct, but I don't have a torch environment in this sandbox to run pytest directly — please let CI (or a manual run) confirm.

jiagaoxiang:
**TestIntel PR Steward:** CI's `code-lint (3.12)` job failed on `323b8616` because `black` (run via pre-commit) wanted to reflow one line in `test_forward_downsamples_spatial_and_channel_dims` (the `AvgDown3D(...)` constructor call was too long) — this is the same line touched by the fix for @WangLingxun's / Copilot's shape-only-assertion finding. I applied exactly the formatting pre-commit proposed (wrapped the constructor call onto multiple lines, no logic change) in `323b8616c024cc38b6a19845e5a566b139640dae` and pushed it to this branch. Expecting `code-lint (3.12)` to go green on the next run.

@WangLingxun / Copilot: the underlying test-content change (full-tensor comparison against an independent reference instead of shape-only) is unchanged from my earlier reply — only formatting moved.

jiagaoxiang:
Superseded by same-repo PR https://github.com/AMD-AGI/Primus/pull/1096 (head `AMD-AGI:testintel/test/cand-9cf47afdeacb6b6d0378129eb4bf01c2`) so Primus-CI-TAS receives Docker Hub credentials and can run torch unit tests. Closing this fork PR.

PR #1087, opened 2026-09-03
test: pin AdaLNContinuous scale/shift order and gradients (PRPUNDIT-17)

jiagaoxiang:

## Summary
- Extends `tests/unit_tests/backends/megatron/diffusion/test_flux_normalization.py` with CPU `TestAdaLNContinuousForwardPlainOps`: independent `norm(x) * (1 + scale) + shift` oracle, zero-modulation LayerNorm reduction, and NeMo chunk order (first half = scale, second half = shift).
- Adds a fused-path CUDA check against the same formula, plus a backward-flow assertion on `x` and `cond`.
- Existing CUDA shape / invalid-norm tests are unchanged.

Closes test gap **PRPUNDIT-17**.

## Test plan
- [ ] GPU CI: `pytest tests/unit_tests/backends/megatron/diffusion/test_flux_normalization.py` (plain-ops CPU class + fused CUDA class; needs `megatron.core`)
- [x] `black --check` / `isort --profile black` on the edited file

### DIFF
--- a/tests/unit_tests/backends/megatron/diffusion/test_flux_normalization.py
+++ b/tests/unit_tests/backends/megatron/diffusion/test_flux_normalization.py
@@ -142,6 +142,106 @@ class TestAdaLNContinuous(PrimusUT):
         with pytest.raises(ValueError, match="Unknown normalization type"):
             AdaLNContinuous(config, conditioning_embedding_dim=HIDDEN_DIM_FLUX, norm_type="invalid")
 
+    def test_fused_forward_matches_plain_ops_formula(self):
+        """Default fused path must still equal norm(x)*(1+scale)+shift (NeMo chunk order)."""
+        config = TransformerConfig(
+            hidden_size=HIDDEN_DIM_FLUX,
+            num_attention_heads=NUM_ATTENTION_HEADS_FLUX,
+            num_layers=1,
+        )
+        adaln = AdaLNContinuous(config, conditioning_embedding_dim=HIDDEN_DIM_FLUX).cuda()
+        torch.manual_seed(0)
+        x = torch.randn(ATTENTION_SEQ_LEN, BATCH_SIZE_QUAD, HIDDEN_DIM_FLUX).cuda()
+        cond = torch.randn(BATCH_SIZE_QUAD, HIDDEN_DIM_FLUX).cuda()
+
+        output = adaln(x, cond)
+        emb = adaln.adaLN_modulation(cond)
+        scale, shift = torch.chunk(emb, 2, dim=1)
+        expected = adaln.norm(x) * (1 + scale) + shift
+        assert torch.allclose(output, expected, atol=1e-5, rtol=1e-5)
+
+
+class TestAdaLNContinuousForwardPlainOps(PrimusUT):
+    """CPU-only tests for AdaLNContinuous.forward numerics.
+
+    These construct AdaLNContinuous with config.adaln_plain_ops=True, which
+    routes forward() through the plain `self.norm(x) * (1 + scale) + shift`
+    branch instead of the primus::fused_ln_modulate custom op (registered
+    only for device_types="cuda"). AdaLNContinuous itself only uses
+    nn.Linear/nn.LayerNorm (no tensor-parallel layers), so no CUDA or
+    Megatron parallel state is required for this path.
+    """
+
+    @staticmethod
+    def _make_config():
+        config = TransformerConfig(
+            hidden_size=HIDDEN_DIM_FLUX,
+            num_attention_heads=NUM_ATTENTION_HEADS_FLUX,
+            num_layers=1,
+        )
+        config.adaln_plain_ops = True
+        return config
+
+    def test_forward_matches_manual_layernorm_and_modulate(self):
+        """forward() should equal norm(x) * (1 + scale) + shift for the given conditioning."""
+        config = self._make_config()
+        adaln = AdaLNContinuous(config, conditioning_embedding_dim=HIDDEN_DIM_FLUX, modulation_bias=False)
+
+        torch.manual_seed(0)
+        x = torch.randn(ATTENTION_SEQ_LEN, BATCH_SIZE_QUAD, HIDDEN_DIM_FLUX)
+        cond = torch.randn(BATCH_SIZE_QUAD, HIDDEN_DIM_FLUX)
+
+        output = adaln(x, cond)
+
+        emb = adaln.adaLN_modulation(cond)
+        scale, shift = torch.chunk(emb, 2, dim=1)
+        expected = adaln.norm(x) * (1 + scale) + shift
+
+        assert output.shape == x.shape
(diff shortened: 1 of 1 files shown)


tests/unit_tests/backends/megatron/diffusion/test_flux_normalization.py
```diff
@@ -142,6 +142,125 @@ def test_invalid_norm_type(self):
...
+        adaln = AdaLNContinuous(config, conditioning_embedding_dim=HIDDEN_DIM_FLUX).cuda()
+        torch.manual_seed(0)
+        x = torch.randn(ATTENTION_SEQ_LEN, BATCH_SIZE_QUAD, HIDDEN_DIM_FLUX).cuda()
+        cond = torch.randn(BATCH_SIZE_QUAD, HIDDEN_DIM_FLUX).cuda()
+
+        output = adaln(x, cond)
+        emb = adaln.adaLN_modulation(cond)
+        scale, shift = torch.chunk(emb, 2, dim=1)
+        expected = adaln.norm(x) * (1 + scale) + shift
+        assert torch.allclose(output, expected, atol=1e-5, rtol=1e-5)
```
jiagaoxiang, line 163: **TestIntel PR Steward:** Good catch — taken. Added `test_fused_forward_backward_reaches_input_and_conditioning` to `TestAdaLNContinuous` (pushed in a4b9621), which runs `.sum().backward()` through the fused CUDA `primus::fused_ln_modulate` path and asserts finite, nonzero gradients reach both `x` and `cond`, mirroring the existing CPU plain-ops backward test.

jiagaoxiang (commented):
Addressed the Copilot finding inline; see the thread reply for details.

jiagaoxiang:
**TestIntel PR Steward:** Follow-up on this poll:

1. **Copilot review comment** (fused-path gradients not covered): addressed. Pushed `a4b9621`, which adds `TestAdaLNContinuous.test_fused_forward_backward_reaches_input_and_conditioning` — a CUDA backward-flow assertion through the `primus::fused_ln_modulate` custom op, mirroring the existing CPU plain-ops backward test. Replied on the thread with details.

2. **`build-docker (3.12)` CI failure**: not caused by this PR. The job builds the image fine and then fails at `docker login -u tasimage -p ${{ secrets.PRIMUS_DOCKER_HUB_TOKEN }}` with `flag needs an argument: 'p' in -p` — i.e. the secret resolves empty. This is expected behavior for `pull_request` runs triggered from a fork (GitHub doesn't expose repo secrets to fork PR workflows), and I see the same `build-docker` failure on every other open `testgap/*` PR right now, so this is a repo/workflow-level CI configuration issue, not something introduced by this test-only change. Rerunning won't help (it's not a flake — the secret will still be empty), and `.github/workflows/ci.yaml` is out of scope for a test PR, so I'm leaving it for a human to fix (e.g. skip the Docker Hub push step for fork-originated `pull_request` runs). Flagging so a maintainer can take a look; not merging until it's resolved and a human approves.

jiagaoxiang:
**TestIntel PR Steward:** Follow-up on this poll:

- Copilot's second review (18:04) flagged three nits, folded into its "suppressed comments" summary rather than posted as new inline threads: (1)/(2) `torch.manual_seed(0)` running after `AdaLNContinuous(...)` construction in two tests, so module init wasn't covered by the seed, and (3) the zero-modulation LayerNorm oracle should use an independent `F.layer_norm` rather than `adaln.norm(x)`. Checking the current head, (3) was already independent (`torch.nn.functional.layer_norm`), and (1)/(2) had already been fixed by moving the seed calls before construction (commit `899c316`).
- However, that same edit (and its follow-up `bf5ed50`) left the file broken: a stray literal `213` line and a duplicated `output = adaln(x, cond)` call in `test_forward_matches_manual_layernorm_and_modulate` — a leftover from a miscounted line-range edit, which would have been a `SyntaxError`/dead code at collection time. I removed both in `676f83e` and confirmed the file now parses cleanly (`ast.parse` on the raw blob).
- Net effect: PR head is now `676f83e`, single non-duplicated call, all three Copilot nits addressed, no functional/test changes beyond the cleanup.

Still holding off on merge: `mergeable_state` is `blocked` (no human approval yet), and the `build-docker (3.12)` job is still expected to fail for this fork PR (empty `PRIMUS_DOCKER_HUB_TOKEN` secret on fork-triggered `pull_request` runs — a repo/workflow config issue out of scope for this test-only change, as flagged earlier). Will merge once a human approves and the in-scope checks are green.

jiagaoxiang:
**TestIntel PR Steward:** Follow-up on this poll:

- Copilot's latest review (18:16, review #5105322135) flagged that `.sum().backward()` in both backward-flow tests (`test_fused_forward_backward_reaches_input_and_conditioning` and `test_forward_backward_reaches_input_and_modulation`) can yield an exactly-zero gradient under LayerNorm's zero-mean-per-token property if modulation ever becomes zero/constant (e.g. a hypothetical future zero-init), making the assertions fragile. This was folded into Copilot's summary rather than posted as inline threads, so replying here instead.
- Addressed: pushed `2263a45`, switching both backward tests to `.pow(2).sum().backward()`, which cannot cancel to zero unless every output element is exactly zero — same intent, more robust loss, no other changes.
- Still holding off on merge: no human approval yet (`mergeable_state` was `blocked`), and `build-docker (3.12)` is still expected to fail on this fork PR for the reason already flagged (empty `PRIMUS_DOCKER_HUB_TOKEN` secret on fork-triggered `pull_request` runs — repo/workflow config issue, out of scope for this test-only change). Will merge once a human approves and in-scope checks are green.

jiagaoxiang:
**TestIntel PR Steward:** Follow-up on this poll:

- Copilot's latest review (18:26, review #5105410572) flagged that the fused CUDA tests (`test_fused_forward_matches_plain_ops_formula`, `test_fused_forward_backward_reaches_input_and_conditioning`) never asserted the module was actually dispatching to the fused `primus::fused_ln_modulate` custom op — so a future flip of the default to the plain-ops branch would silently stop exercising the fused kernel while these tests kept passing.
- Addressed: pushed `70e8655`, adding `assert adaln.use_fused_ln_modulate, "expected fused CUDA dispatch (adaln_plain_ops=False)"` right after construction in both fused tests. No other changes.
- Still holding off on merge: `mergeable_state` is `blocked` (no human approval yet), and `build-docker (3.12)` is still expected to fail on this fork PR for the reason flagged in earlier follow-ups (empty `PRIMUS_DOCKER_HUB_TOKEN` secret on fork-triggered `pull_request` runs — a repo/workflow config issue out of scope for this test-only change, and observed on every other open `testgap/*` PR). Will merge once a human approves and the in-scope checks are green.

jiagaoxiang:
Superseded by same-repo PR https://github.com/AMD-AGI/Primus/pull/1100 (head `AMD-AGI:testgap/PRPUNDIT-17`) so Primus-CI-TAS receives Docker Hub credentials and can run torch unit tests. Closing this fork PR.

PR #1092, opened 2026-09-03
test: pin AttentionResidualMixer against an independent mix (PRPUNDIT-22)

jiagaoxiang:

## Summary
- Adds `tests/unit_tests/backends/megatron/kimi_k3/test_attention_residual.py`, the file the module docstring already claimed existed.
- Independent candidate-by-candidate RMS-score / unnormalised-value oracle (not a copy of the production eager kernel), plus `num_blocks=0` identity, hidden-size `ValueError`, dtype cast-back, and `gradcheck`.

Closes test gap **PRPUNDIT-22**.

## Test plan
- [ ] CI with `megatron.core`: `pytest tests/unit_tests/backends/megatron/kimi_k3/test_attention_residual.py` (skipped on this host; `megatron.core` not importable)
- [x] `black --check` / `isort --profile black` on the new files

### DIFF
--- a/tests/unit_tests/backends/megatron/kimi_k3/test_attention_residual.py
+++ b/tests/unit_tests/backends/megatron/kimi_k3/test_attention_residual.py
@@ -1,0 +1,230 @@
+###############################################################################
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+#
+# See LICENSE for license information.
+###############################################################################
+
+"""Tests for ``AttentionResidualMixer.forward``.
+
+The mixer replaces ``x = x + sublayer(x)`` with a learned softmax mixture over
+the running residual stream (``prefix_sum``) and a set of cross-layer
+checkpoints (``block_residual``); see the module docstring at
+``primus/backends/megatron/core/transformer/kimi_k3/attention_residual.py``
+for the three easy-to-get-wrong details this file pins:
+
+1. the output mixes the *un-normalised* candidates -- RMS normalisation feeds
+   the softmax scores only;
+2. the scorer is rank-1, ``norm_weight (elementwise*) proj_weight``;
+3. everything runs in fp32 (or wider) internally and casts back to
+   ``prefix_sum``'s dtype at the end.
+
+``_reference_attn_res_mix`` below is an independent transcription of
+``_apply_attn_res`` (``modeling_kimi_linear.py``) -- written without looking at
+``attn_res_kernels._eager.reference.eager_attn_res_mix`` -- so a regression in
+either implementation has a real chance of being caught rather than the test
+merely checking the module against its own backend.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+torch = pytest.importorskip("torch")
+
+mcore_module = pytest.importorskip(
+    "megatron.core.transformer.module",
+    reason="megatron.core is not importable in this environment",
+)
+
+from megatron.core.transformer.transformer_config import TransformerConfig  # noqa: E402
+
+from primus.backends.megatron.core.transformer.kimi_k3.attention_residual import (  # noqa: E402
+    AttentionResidualMixer,
+)
+
+
+def _make_config(hidden_size=8, eps=1e-6, params_dtype=torch.float32, attn_res_backend="eager"):
+    config = TransformerConfig(
+        hidden_size=hidden_size,
+        num_attention_heads=1,
+        num_layers=1,
+        layernorm_epsilon=eps,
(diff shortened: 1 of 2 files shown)


tests/unit_tests/backends/megatron/kimi_k3/test_attention_residual.py
```diff
@@ -0,0 +1,230 @@
...
+from __future__ import annotations
+
+import pytest
+
+torch = pytest.importorskip("torch")
+
+pytest.importorskip(
+    "megatron.core.transformer.module",
+    reason="megatron.core is not importable in this environment",
+)
```
jiagaoxiang, line 37: TestIntel PR Steward: taken — fixed in 529ccbf2b1730b55bb61b82a11370acc5dce88e9 by calling `pytest.importorskip(...)` without binding the result, matching your suggestion.

tests/unit_tests/backends/megatron/kimi_k3/test_attention_residual.py
```diff
@@ -0,0 +1,230 @@
...
+    mixer.proj_weight.requires_grad_(True)
+
+    # norm_weight / proj_weight are passed through so gradcheck perturbs the
+    # exact tensors the mixer reads (they are the same objects, not copies).
+    assert torch.autograd.gradcheck(
+        lambda ps, br, nw, pw: mixer(ps, br),
+        (prefix_sum, block_residual, mixer.norm_weight, mixer.proj_weight),
+        eps=1e-6,
+        atol=1e-4,
+    )
```
jiagaoxiang, line 230: TestIntel PR Steward: skipping this one — the lambda's parameter list ignores `nw`/`pw`, but the tuple passed to `gradcheck` is `(prefix_sum, block_residual, mixer.norm_weight, mixer.proj_weight)`, i.e. the *exact same tensor objects* the mixer reads internally (see the comment directly above the assert). `gradcheck`'s numerical Jacobian perturbs each input tensor's storage in place, and its analytical Jacobian calls `autograd.grad(output, inputs)` against those same leaf objects — both work correctly by object identity even though the lambda's own argument names go unused. So the check does validate gradients w.r.t. `norm_weight`/`proj_weight` today. Happy to revisit if you have a concrete case where this breaks, but I don't want to change working gradient-check plumbing without one.

jiagaoxiang (commented):
TestIntel PR Steward: addressed the Copilot review findings — see inline replies.

jiagaoxiang:
**TestIntel PR Steward:** follow-up on this run's findings.

**Copilot review (2 comments)**
- Unused `mcore_module` binding: fixed — pushed 529ccbf2b1730b55bb61b82a11370acc5dce88e9, which drops the unused assignment and calls `pytest.importorskip(...)` purely for its skip side effect.
- `gradcheck` lambda "ignoring" `nw`/`pw`: declined — replied inline. The tuple passed to `gradcheck` contains the exact same tensor objects (`mixer.norm_weight`, `mixer.proj_weight`) the mixer reads internally, so both `gradcheck`'s numerical and analytical Jacobians correctly exercise those parameters despite the lambda not using its own argument names. Changing this wouldn't fix a real bug and risks weakening a currently-correct check.

**CI — `build-docker (3.12)` failure**
Looked at the failed job (run 33780529743): the actual Docker build/image steps complete successfully; the job fails a step later at `docker login` with `flag needs an argument: 'p' in -p`, i.e. the registry password argument is empty. `run-unittest-torch` (the job that would actually execute this PR's new test) is gated behind `build-docker` and got skipped as a result — it never ran.

This isn't caused by anything in this PR (only test files were touched) and isn't a plain flake either — it reads like the Docker Hub push step's PAT secret isn't available in this run's context (this PR runs from a fork branch), which is standard GitHub Actions behavior for `pull_request`-triggered jobs from forks and would reproduce on any fork PR the same way. Since I can't safely fix repo secrets/workflow trust configuration from here (and doing so is outside this test PR's scope), I'm not rerunning the job or touching the workflow — flagging for a maintainer with repo-admin access to either re-trigger this run in a context with the secret available, or confirm `run-unittest-torch` can be validated another way.

I'll keep watching for review activity and CI updates, and will merge once there's a human approval and green (or maintainer-cleared) required checks.

jiagaoxiang:
Superseded by same-repo PR https://github.com/AMD-AGI/Primus/pull/1105 (head `AMD-AGI:testgap/PRPUNDIT-22`) so Primus-CI-TAS receives Docker Hub credentials and can run torch unit tests. Closing this fork PR.

PR #1093, opened 2026-09-03
test: cover fused clamped SwiGLU Triton autograd (PRPUNDIT-23)

jiagaoxiang:

## Summary
- Adds `tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py` for the production Triton wrappers in `fused_bias_swiglu.py`.
- One GPU-gated suite covers unweighted `ClampedSwiGLUFunction` and weighted `ClampedWeightedSwiGLUFunction` (forward, backward, `fp8_input_store`, impl routing) against an eager clamp/silu/mul reference.
- Per linxwang: do **not** dump this into eager `test_clamped_swiglu.py`; the weighted path is dormant in shipped V4 config but stays in this shared suite. Weighted is not a second ticket.

Closes test gap **PRPUNDIT-23**.

## Test plan
- [ ] GPU CI: `pytest tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py` (skipped on this host; no CUDA)
- [x] `black --check` / `isort --profile black` on the new file

### DIFF
--- a/tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py
+++ b/tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py
@@ -26,8 +26,14 @@ cuda = pytest.mark.skipif(not torch.cuda.is_available(), reason="clamped SwiGLU 
 def _fp8_cast_works() -> bool:
     if not hasattr(torch, "float8_e4m3fn"):
         return False
+    # This suite only ever runs its parametrized cases on CUDA/HIP (the
+    # module is skipped entirely otherwise via the `cuda` mark below), so
+    # probe the cast on the device the kernels actually execute on. A
+    # CPU-only probe can incorrectly skip fp8_input_store=True on builds
+    # where FP8 casting is CUDA-only.
+    device = "cuda" if torch.cuda.is_available() else "cpu"
     try:
-        torch.zeros(4).to(torch.float8_e4m3fn).to(torch.float32)
+        torch.zeros(4, device=device).to(torch.float8_e4m3fn).to(torch.float32)
         return True
     except (RuntimeError, TypeError):
         return False
--- a/tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py
+++ b/tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py
@@ -1,0 +1,121 @@
+###############################################################################
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+#
+# See LICENSE for license information.
+###############################################################################
+
+"""Parameterized Triton clamped-SwiGLU autograd tests (PRPUNDIT-23).
+
+``test_clamped_swiglu.py`` covers the eager ``clamped_swiglu.py`` module.
+This file covers the production Triton wrappers in ``fused_bias_swiglu.py``:
+unweighted ``ClampedSwiGLUFunction`` (shared-expert MLP) and weighted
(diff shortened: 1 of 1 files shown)


jiagaoxiang:
**TestIntel PR Steward:** Follow-up on this poll:

**CI (`build-docker (3.12)`, run 33780535897) — unrelated repo-wide breakage, not caused by this PR.** The docker image itself built fine; the job then failed at the `docker login` step with `flag needs an argument: 'p' in -p` / exit code 125, i.e. a malformed/missing registry-password argument on the runner. This hit every `testgap/*` PR created in the same window (PRPUNDIT-17, 18, 20, 21, 22, and `compressor-forward`) at the same timestamp, so it's an infra/credentials issue on the base branch's CI, not something this test-only diff can fix. Not rerunning (it's not a transient flake — it's a consistently broken login step) and not merging until a human resolves the runner/registry credentials and CI goes green.

**Copilot review (`tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py#discussion_r3926704703`) — addressed.** Copilot correctly flagged that `fp8_input_store=True` parametrized cases can hard-fail on GPUs/torch builds without a working `float8_e4m3fn` cast, even with CUDA available. Pushed `400980c`: added an `_fp8_cast_works()` probe (mirroring `test_v4_fp8_indexer.py`) and used it to skip only the `fp8_input_store=True` parameter via `pytest.param(..., marks=...)`, so `fp8_input_store=False` coverage keeps running on non-FP8 devices instead of the whole test hard-failing. Ran `black`/`isort` (project config: line-length 110) on the changed file — clean.

I'll keep following this PR for CI status, further review comments, and merge-readiness (open, green required checks, human approval, no outstanding change requests).

jiagaoxiang:
**TestIntel PR Steward:** Follow-up on this poll (new Copilot review + persisting CI status).

**Copilot review (`test_fused_clamped_swiglu.py#discussion_r3927358090`, line 33) — addressed.** Copilot correctly flagged that `_fp8_cast_works()` probed the `float8_e4m3fn` cast on a CPU tensor even though every parametrized case in this suite only ever runs on `device="cuda"` (the whole module is skipped otherwise via the `cuda` mark). On a build where FP8 casting is CUDA-only, that CPU probe would incorrectly report "unsupported" and skip valid `fp8_input_store=True` coverage. Pushed `e448f2e` to probe on `cuda` when available (falling back to `cpu` only if CUDA isn't present), so the gate now reflects the device the kernels actually execute on.

**Copilot review (same review, summary text) — backward `fp8_input_store=True` coverage: declining for now.** Copilot suggested parametrizing `test_backward_matches_eager_autograd` over `fp8_input_store` too. Unlike the forward tests, there's no clean eager oracle for the FP8-storage backward path — the input is quantized to `float8_e4m3fn` and reconstructed before backward, so an exact `assert_close` against the fp32 eager backward isn't valid, and "assert grads are finite" is a much weaker check than the rest of this suite uses. Adding that would need a considered numerical-tolerance design, which is beyond this test-gap ticket (PRPUNDIT-23 scoped to matching eager behavior for the existing forward/backward/routing cases). Leaving this as a follow-up rather than bolting on a weaker assertion style to this PR.

**CI (`build-docker (3.12)`) — still the same unrelated infra breakage** noted in my last comment (broken `docker login` step on the runner, hitting every `testgap/*` PR in this window). No new failure introduced by `e448f2e`; not rerunning, not merging until CI is green and a human approves.

Still watching for CI, further review comments, and merge-readiness.

jiagaoxiang:
Superseded by same-repo PR https://github.com/AMD-AGI/Primus/pull/1106 (head `AMD-AGI:testgap/PRPUNDIT-23`) so Primus-CI-TAS receives Docker Hub credentials and can run torch unit tests. Closing this fork PR.

PR #1097, opened 2026-09-04
testintel: add missing test for Decoder3d.forward (created by linxwang)

jiagaoxiang:
Same-repo refile of #1084 so Primus-CI-TAS runs against `AMD-AGI/Primus` (fork PRs do not receive Docker Hub credentials, so `build-docker` / torch unit tests never ran).

The feat_cache/feat_idx chunked-decode bookkeeping in Decoder3d.forward (and the CausalConv3d/Resample/ResidualBlock layers it drives) had no test coverage, leaving index-drift or cache-shape regressions undetected.

This PR adds a focused unit test for `Decoder3d.forward` covering the real per-frame `WanVAE_.decode` protocol: a shared `feat_cache`, `feat_idx` reset per frame, `first_chunk` handling, cache-slot/sentinel transitions (the Rep marker), temporal output counts, and deterministic replay after a cache reset. A no-cache/single-shot decode is not a valid oracle here: `Resample.upsample3d` only performs temporal doubling inside the feat_cache-is-not-None branch, so there is no equivalent non-streaming decode path to compare against (see the review-thread discussion). This PR does not change production code.


tests/unit_tests/backends/diffusion/test_wan_vae_decoder.py
```diff
@@ -0,0 +1,116 @@
...
+def _make_decoder():
+    # A tiny decoder with a single temporal-upsample stage: enough to exercise
+    # cache indexing, the "Rep" sentinel, and first_chunk propagation without
+    # a channel-changing residual shortcut (which vae2_2's ResidualBlock does
+    # not route through the feat_cache, and would desync feat_idx from
+    # count_conv3d(decoder)).
+    # Seed only within a forked RNG scope so this doesn't mutate global RNG
+    # state and leak into other tests run later in the same process.
+    with torch.random.fork_rng(devices=[]):
+        torch.manual_seed(0)
```
jiagaoxiang, line 31: TestIntel PR Steward: Good catch — fixed in 8204f01 by passing `devices=[]` to `torch.random.fork_rng()`, so the seeding scope no longer forks CUDA RNG state on GPU-enabled runners.

tests/unit_tests/backends/diffusion/test_wan_vae_decoder.py
```diff
@@ -0,0 +1,116 @@
...
+    # First chunk has no cached history yet, so its upsample3d stage can only
+    # emit its own frame; later chunks have history and double their
+    # temporal contribution.
+    assert outputs[0].shape == (1, 12, 1, 8, 8)
+    assert outputs[1].shape == (1, 12, 2, 8, 8)
+    assert outputs[2].shape == (1, 12, 2, 8, 8)
+
+    out = torch.cat(outputs, dim=2)
+    assert out.shape == (1, 12, 5, 8, 8)
+    assert torch.isfinite(out).all()
```
jiagaoxiang, line 118: TestIntel PR Steward: Declining this one after checking the architecture in `vae2_2.py`. `Resample.forward` (mode `upsample3d`) only performs the temporal 2x doubling inside the `if feat_cache is not None:` branch — when `feat_cache` is `None` it falls through to spatial-only upsampling and the time dimension is left unchanged. So there isn't a "single-shot, no-cache" decode path in this model that reproduces the streamed result: `WanVAE_.decode` itself always drives `Decoder3d` frame-by-frame through a shared `feat_cache`/`feat_idx`, exactly like this test does. A full-sequence call with a fresh cache and `first_chunk=True` also isn't equivalent, since the temporal-upsample branch only skips doubling on the very first cache slot access regardless of how many frames are in that call, so it wouldn't reproduce the expected 5-frame streamed output either. Given that, the cache-bookkeeping + determinism assertions already in this PR are the correct/only feasible way to pin down this behavior; a "no-cache reference" comparison would just be comparing against a different, non-equivalent computation. Happy to revisit if you see a concrete construction that sidesteps this.

jiagaoxiang (commented):
TestIntel PR Steward: Replies to Copilot's two findings inline.

jiagaoxiang:
**TestIntel PR Steward:** Addressed the Copilot review from this run (new head `8204f01`):

- RNG fork: fixed — `torch.random.fork_rng(devices=[])` now avoids forking CUDA RNG state in this CPU-only test.
- Numeric streaming-vs-single-shot parity: declined, with rationale posted inline. `Resample.forward`'s `upsample3d` path only performs temporal doubling inside the `feat_cache is not None` branch, so there's no "no-cache single-shot" decode in this model to compare against — `WanVAE_.decode` itself always drives the decoder frame-by-frame through a shared cache, which is exactly what the existing test already exercises.

This PR is still draft, so no merge action taken. Will keep following CI, reviews, and mergeability.

WangLingxun:
The revised test code now follows the actual Wan 2.2 production protocol and no longer uses a no-cache/single-shot decode as an oracle, which is correct. However, the PR body still says the gap is that streaming/chunked decode does not verify reproduction of a single-shot output. That acceptance criterion is invalid for this decoder: the no-cache path does not execute the same temporal-upsample/cache protocol and can have different temporal behavior.

Please update the PR description before marking it ready. It should instead say that the test covers the real per-frame `WanVAE_.decode` protocol: shared `feat_cache`, `feat_idx` reset per frame, `first_chunk` handling, cache-slot/sentinel transitions, temporal output counts, and deterministic replay after cache reset. No test-code change is requested here.

jiagaoxiang:
**TestIntel PR Steward:** Updated the PR description per @WangLingxun's comment — it no longer claims streaming/chunked decode reproduces a single-shot decode output (that oracle is invalid for this decoder, as discussed in the review thread). The description now states the actual coverage: the per-frame `WanVAE_.decode` protocol — shared `feat_cache`, `feat_idx` reset per frame, `first_chunk` handling, cache-slot/sentinel transitions, temporal output counts, and deterministic replay after a cache reset. No test-code change was requested or made.

Still draft, `mergeable_state: behind` main, and no failing checks. Will keep following.

tests/unit_tests/backends/diffusion/test_wan_vae_decoder.py
```diff
@@ -0,0 +1,116 @@
+##########################################################################
+# Copyright (c) 2025, Advanced Micro Devices, Inc.
```
WangLingxun: This is a newly added file, so the copyright year should be 2026.

jiagaoxiang:
**TestIntel PR Steward:** Addressed @WangLingxun's review comment on `tests/unit_tests/backends/diffusion/test_wan_vae_decoder.py:2` — the copyright header now reads 2026 instead of 2025, since this is a newly added file. Pushed as `bb2afaa`.

No other new findings from this poll. Still draft; `mergeable_state` was `behind` main but no failing checks. Will keep following CI, reviews, and mergeability.

jiagaoxiang:
**TestIntel PR Steward:** Addressed the latest Copilot review (pullrequestreview-5128066056, on `bb2afaa`): both tests still generated the decoder input `x = torch.randn(1, 4, 3, 4, 4)` outside the forked RNG scope, which advanced global RNG state despite the earlier fix that scoped decoder-init seeding. Fixed by factoring input generation into a `_make_input()` helper that draws `x` inside `torch.random.fork_rng(devices=[])` with a fixed seed, mirroring the existing `_make_decoder()` pattern. Pushed as `a7805ce`.

Still draft, no failing checks (only the Read the Docs status check, which is green). Will keep following CI, reviews, and mergeability.

jiagaoxiang:
**TestIntel PR Steward:** Addressed the latest Copilot review comment on `tests/unit_tests/backends/diffusion/test_wan_vae_decoder.py:20` (discussion r3946578926, pullrequestreview-5128094912): `torch.manual_seed()` also seeds CUDA/ROCm RNG state when a GPU runtime is present, so `fork_rng(devices=[])` wasn't restoring it afterward, which could leak into other tests on GPU-capable runners — as flagged, this also applied to the `_make_input()` call at line ~46.

Fixed by adding a `_fork_rng_devices()` helper that forks `range(torch.cuda.device_count())` when CUDA is available and `[]` when it isn't, and using it in both `_make_decoder()` and `_make_input()`. This keeps the original intent (no CUDA runtime init forced on CPU-only runners, per the first Copilot comment on this same line) while properly scoping/restoring CUDA RNG state when a GPU is present. Pushed as `3bf4791`.

Still draft, no failing checks (only the Read the Docs status, which is green). `mergeable_state` was `behind` main but that's a metadata field, not a conflict — GitHub reports `mergeable: true`. Will keep following CI, reviews, and mergeability.

Xiaoming-AMD (approved):

PR #1106, opened 2026-09-04
test: cover fused clamped SwiGLU Triton autograd (PRPUNDIT-23)

jiagaoxiang:
Same-repo refile of #1093 so Primus-CI-TAS runs against `AMD-AGI/Primus` (fork PRs do not receive Docker Hub credentials, so `build-docker` / torch unit tests never ran).

## Summary
- Adds `tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py` for the production Triton wrappers in `fused_bias_swiglu.py`.
- One GPU-gated suite covers unweighted `ClampedSwiGLUFunction` and weighted `ClampedWeightedSwiGLUFunction` (forward, backward, `fp8_input_store`, impl routing) against an eager clamp/silu/mul reference.
- Per linxwang: do **not** dump this into eager `test_clamped_swiglu.py`; the weighted path is dormant in shipped V4 config but stays in this shared suite. Weighted is not a second ticket.

Closes test gap **PRPUNDIT-23**.

## Test plan
- [ ] GPU CI: `pytest tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py` (skipped on this host; no CUDA)
- [x] `black --check` / `isort --profile black` on the new file

### DIFF
--- a/tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py
+++ b/tests/unit_tests/megatron/transformer/deepseek_v4/test_fused_clamped_swiglu.py
@@ -82,15 +82,25 @@ class TestClampedSwiGLUFunction:
         y = (torch.randn(17, 64, device="cuda", dtype=torch.float32) * 5.0).requires_grad_(True)
         out = self.Fn.apply(y, fp8_input_store, alpha)
         torch.testing.assert_close(out, _eager_clamped_swiglu(y, alpha), atol=1e-5, rtol=1e-5)
+        (saved_input,) = out.grad_fn.saved_tensors
+        assert saved_input.dtype == (torch.float8_e4m3fn if fp8_input_store else y.dtype)
 
-    def test_backward_matches_eager_autograd(self):
+    @pytest.mark.parametrize("fp8_input_store", _fp8_input_store_cases)
+    def test_backward_matches_eager_autograd(self, fp8_input_store: bool):
         torch.manual_seed(6)
         alpha = 7.0
         y_fn = (torch.randn(5, 20, device="cuda", dtype=torch.float32) * 5.0).requires_grad_(True)
-        y_eager = y_fn.detach().clone().requires_grad_(True)
+        y_eager = y_fn.detach().clone()
+        if fp8_input_store:
+            # Backward recomputes through the FP8-quantized value forward saved
+            # for the input, not the full-precision input; mirror that lossy
+            # round trip here so the reference exercises the same path instead
+            # of silently comparing against full precision.
+            y_eager = y_eager.to(torch.float8_e4m3fn).to(torch.float32)
+        y_eager = y_eager.requires_grad_(True)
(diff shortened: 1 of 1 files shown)


jiagaoxiang:
**TestIntel PR Steward:** Addressed the Copilot review finding on `test_fused_clamped_swiglu.py` (thread on line 84, also flagged for lines 92/139) in `3638364`:

- `fp8_input_store` was parametrized for forward, but only the output was checked against the eager reference — the FP8-save/restore path in `backward` was never actually exercised, and the unweighted forward test never asserted which dtype got saved (unlike the weighted one).
- `TestClampedSwiGLUFunction.test_forward_matches_eager_reference` now asserts `out.grad_fn.saved_tensors[0].dtype` matches the flag, mirroring the weighted test.
- Both `test_backward_matches_eager_autograd` tests are now parametrized over `fp8_input_store`. When `True`, the eager reference dequantizes through `float8_e4m3fn` first — the same lossy round trip `ClampedSwiGLUFunction.backward` / `ClampedWeightedSwiGLUFunction.backward` perform — so the comparison stays at the existing tight `atol/rtol=1e-5` instead of masking the FP8 path behind a loosened tolerance.

Also for the record on CI: the `build-docker (3.12)` failure the poller saw was from a stale run on the pre-refile fork PR (no Docker Hub credentials there, as noted in the PR description). The current same-repo run's `build-docker (3.12)` job has already passed; `run-unittest-jax` / `run-unittest-torch` are still in progress. No rerun needed.

Will keep watching CI and for a human approval before merging.

jiagaoxiang:
**TestIntel PR Steward:** Follow-up on the second Copilot review ([pullrequestreview-5115815432](https://github.com/AMD-AGI/Primus/pull/1106#pullrequestreview-5115815432), submitted 16:58 UTC, commit `3638364`).

That review suggests extending the new parity suite from fp32-only to also cover bf16 (and optionally fp16) for both the unweighted and weighted forward/backward tests, since the production kernels run in mixed precision.

Declining for this PR, scope-only: PRPUNDIT-23's gap was untested clamped-SwiGLU autograd *correctness* against the eager reference — it didn't call for dtype-precision coverage across the training dtype matrix. Adding bf16/fp16 would double the parametrization surface and need new dtype-aware tolerances, which is a materially different (and larger) ask than what was scoped here. That's consistent with the scope line already in the PR description ("weighted is not a second ticket") — same principle applies to dtype expansion. If mixed-precision parity for this kernel is wanted, it should be its own follow-up ticket rather than folded into this gap-closure PR.

No code change from this feedback (the earlier, first Copilot review — fp8_input_store not exercised through backward — was already addressed in `3638364`, see prior comment on this PR).

Status: `run-unittest-jax` / `run-unittest-torch` were still pending as of this poll; `build-docker (3.12)` and the Read the Docs check are green. No human review yet. Will keep watching and merge only after a human approval plus green required checks.

Xiaoming-AMD (approved):

PR #1116, opened 2026-09-08
fix(megatron): checkpoint heterogeneous Flux layers independently

gphuang:
## Summary

Flux combines joint and single transformer blocks whose AdaLN tensors have incompatible shapes. The MXFP6 branch currently uses Megatron's homogeneous layer-stacked checkpoint namespace, so its first distributed checkpoint save aborts before writing any files.

- Require `FluxConfig.hetereogenous_dist_checkpoint=True`, which gives each Flux layer its own checkpoint key.
- Reject the invalid homogeneous layout instead of deferring failure until a long run reaches its first save.
- Add coverage for distinct joint/single AdaLN keys and a real `torch_dist` save/load round trip.
- Format files on the consolidated MXFP6 branch that were blocking the repository-wide lint check.

## Test plan

- [x] `py_compile` passes for all changed checkpoint files.
- [x] `isort --check-only` and `black --check` pass across all Primus-owned Python files.
- [x] The same checkpoint fix was previously reviewed and merged on another Flux branch.
- [x] A full Flux 12B prefix save and resumed continuation previously validated that merged fix.
- [ ] Required CI checks pass on this MXFP6 branch.
- [ ] Run a minimal save/load/resume smoke from the merged branch before restarting the long prefix.

### DIFF
--- a/primus/backends/megatron/core/extensions/primus_turbo_mxfp6_local.py
+++ b/primus/backends/megatron/core/extensions/primus_turbo_mxfp6_local.py
@@ -47,7 +47,10 @@ from primus_turbo.pytorch.core.low_precision import (
     MXFP6_PROLOGUE_BIAS_GELU_BACKWARD,
     ScalingGranularity,
 )
-from primus_turbo.pytorch.kernels.gemm.gemm_fp6_impl import gemm_fp6_impl, gemm_fp6_out_impl
+from primus_turbo.pytorch.kernels.gemm.gemm_fp6_impl import (
+    gemm_fp6_impl,
+    gemm_fp6_out_impl,
+)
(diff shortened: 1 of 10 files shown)


gphuang:
## Verdict: heterogeneous Flux checkpoint fix is ready to merge

**No blocking findings.** The patch selects Megatron's supported per-layer namespace and keeps checkpoint integrity validation strict; it does not hide the AdaLN shape conflict with `allow_shape_mismatch`.

Reviewed the complete current-base diff. Syntax and repository-wide formatter checks pass, required CI is green, the regression covers distinct joint/single AdaLN keys plus a real `torch_dist` round trip, and the same fix previously completed a full Flux 12B checkpoint save and resumed continuation.

PR #1117, opened 2026-09-08
test(flux): use unittest-compatible checkpoint tempdir

gphuang:
## Summary

The Flux distributed-checkpoint round-trip test was added to a `unittest.TestCase` subclass but requested pytest's `tmp_path` fixture. Pytest therefore called the method without that argument and failed before exercising checkpoint I/O.

- Replace the fixture argument with a scoped `TemporaryDirectory`.
- Keep the checkpoint directory alive through save, load, and tensor comparison.

## Test plan

- [x] `isort --check-only`
- [x] `black --check`
- [x] `py_compile`
- [x] Run the corrected distributed-checkpoint tests in the v26.5 development container (12 passed).

### DIFF
--- a/tests/unit_tests/backends/megatron/diffusion/test_flux_dist_checkpoint.py
+++ b/tests/unit_tests/backends/megatron/diffusion/test_flux_dist_checkpoint.py
@@ -3,6 +3,9 @@
 
 """Distributed-checkpoint regression tests for Flux's heterogeneous layers."""
 
+import tempfile
+from pathlib import Path
+
 import pytest
 import torch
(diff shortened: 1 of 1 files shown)


gphuang:
## Note: author review verdict

No blocking findings. The change removes the unsupported pytest fixture from the `unittest.TestCase` method and keeps the temporary checkpoint alive through save, load, and tensor comparison.

Static formatting checks pass, CI lint is green, and the corrected checkpoint suite passes 12/12 in the v26.5 development container. Ready to merge.

PR #1165, opened 2026-09-16
feat: add fused grouped gemm support

RuibinCheung:
# Description

Wires up `turbo_fused_grouped_gemm`, which until now existed only as an inert config switch.

When enabled, `PrimusGroupedMLP` replaces the `fc1 -> weighted GLU -> fc2` chain — three separate Turbo ops with the wide bf16 intermediate materialized between them — with a single Primus-Turbo `grouped_mlp_fp8` / `grouped_mlp_fp4` call that also owns WGRAD.

Getting there required pulling weight preparation out of `PrimusTurboGroupedLinear.forward`: the first-microbatch quantized-weight cache, the MXFP4 de-osc dtype marker and the `_bridge_weight_grad` hop that keeps `main_grad` on the real Parameter are all state the fused op needs, but they were previously inlined in the grouped-GEMM forward and duplicated across its FP8 and FP4 branches. They now live in a shared `prepare_weights()` that both callers use, which also collapses the four near-identical `get_num_microbatches() == 1` / multi-microbatch blocks down to one per format.

The flag is off by default, and every path it touches is unchanged when it is off.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Infra/Build change
- [x] Code refactoring

## Changes

Please list the changes introduced in this PR:

- `experts.py`: add `PrimusGroupedMLP._forward_with_turbo_fused_grouped_mlp()`, dispatching to `grouped_mlp_fp8` or `grouped_mlp_fp4` depending on which Turbo low-precision mode is active, with `activation_recompute` honored via `tensor_parallel.checkpoint`. Weight prep deliberately stays outside the checkpoint so the first-microbatch cache, the de-osc marker and the grad bridge run exactly once.
- `experts.py`: add `is_turbo_fused_grouped_mlp_supported()`. The fused op currently covers Turbo FP8 under current or MXFP8 scaling, and Turbo FP4 under MXFP4 scaling. Anything else — block scaling, or no Turbo quantization at all — falls back to the unfused chain with a warn-once message rather than failing the run.
- `experts.py`: add `_check_turbo_fused_grouped_gemm_contract()`, asserted at `__init__` so unsupported configs fail at construction instead of in the first forward. It rejects `add_bias_linear`, a non-zero `glu_linear_offset`, `activation_func_fp8_input_store`, `moe_apply_probs_on_input` (the fused GLU applies probs after fc1, which is a different contract), fc1 / moe-act offload, and linear layers that are not `PrimusTurboGroupedLinear`.
- `primus_turbo.py`: extract `PrimusTurboGroupedLinear.prepare_weights()` from `forward` / `forward_internal` and reuse it from both the grouped GEMM and the fused MLP. `is_first_microbatch` ownership moves into it, so `forward_internal` no longer takes that argument.
- `primus_turbo.py`: plumb `activation_func_clamp_value` through `fused_bias_act_with_probs` to the Turbo kernel's `clamp_limit`. `use_turbo_fused_act_with_probs` previously asserted the clamp value was `None`; that restriction is lifted.
- `ci.yaml`: bump `PRIMUS_TURBO_COMMIT` to `a26a5066` for the two kernel features this wiring depends on — MXFP8 support in the fused grouped MLP (Primus-Turbo #503) and clamp-limit support in the GLU kernel (#506).
- `primus_turbo.yaml`: drop the "not merged yet, so the flag is inert" note now that the flag does something.
- New test `tests/unit_tests/megatron/moe/test_fused_grouped_gemm.py`.

## Test

`test_fused_grouped_gemm.py` builds two `PrimusGroupedMLP` instances that differ only in the flag, copies one's weights onto the other, and runs both under the same Turbo FP8 autocast over identical tokens, router probs and incoming gradient.

It compares the forward output and all four gradients (dx, dW1, dW2, dprobs) against a naive fp32 per-expert SwiGLU reference, using the same SNR-in-dB metric and the same 20 dB floor that Primus-Turbo's own `tests/pytorch/ops/test_grouped_mlp_fp8.py` applies to this op. The two paths are additionally held to a tighter floor against each other, so a fused implementation that is wrong in a way the unfused chain happens to share still fails the first check.

8 cases: `{tensorwise, mxfp8} x {128-aligned, ragged group lengths} x {1, 2 microbatches}`. The ragged lengths exercise `turbo_grouped_gemm_without_padding`; two microbatches exercise the `main_grad` WGRAD bridge, where the real weight gradients land in `main_grad` rather than `.grad`.

All 8 pass on MI355X. Measured 22.8-23.9 dB against fp32 on every tensor, and 32-52 dB fused vs unfused.

# Checklist:

- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes


RuibinCheung:
PTAL @HuangWei-95

Xiaoming-AMD (approved):

PR #1166, opened 2026-09-16
test(megatron): let the Flux checkpoint round-trip test actually run

gphuang:
## Summary

The regression test added in #1116 to guard the heterogeneous Flux AdaLN checkpoint fix has never executed. `TestFluxDistCheckpoint` extends `PrimusUT`, which is a `unittest.TestCase`, and `test_torch_dist_save_load_round_trip` asks for pytest's `tmp_path` fixture. pytest does not inject fixtures into `TestCase` methods — it calls the method with no arguments, so Python raises `TypeError` before any checkpoint logic runs.

```
TypeError: TestFluxDistCheckpoint.test_torch_dist_save_load_round_trip() missing 1 required positional argument: 'tmp_path'
```

This surfaced on an MI355X node while validating the checkpoint stack for the MXFP6 ablation work: the test fails identically on CPU and on hardware, because the failure happens at call time and never reaches the save/load path.

### Change

Replace the fixture with `tempfile.TemporaryDirectory()`, which works inside a `TestCase`. Assertions are unchanged and no production code is touched. `TestFluxDistCheckpoint` was the only `TestCase` subclass under `tests/` requesting a pytest fixture; the other fixture users (`TestArrowToTarTimestep`, `TestCheckpointConversion`, `TestDiffusionPretrainTrainer`) are plain classes, where injection works normally.

## Test plan

- [x] Failure reproduced and root-caused from a container run on `rocm/primus:v26.5`
- [x] Confirmed `PrimusUT` is a `unittest.TestCase` subclass (`tests/utils.py`)
- [x] Swept `tests/` for the same pattern — no other occurrence
- [ ] **Not yet executed green.** The test needs torch, Megatron, and a GPU, and this branch's CI cannot run unit tests: `code-lint` fails on `feat/mxfp6-fused-mlp` itself, `build-docker` requires `code-lint`, and `run-unittest-torch` requires `build-docker`. The next MI355X session will run it in-container with `pytest` and report back here.

The change is strictly smaller than the status quo: today the test raises `TypeError` unconditionally, so it cannot regress from here.


gphuang:
## Verdict: ready to merge

No blocking findings. Reviewed the full base-to-head diff, not just the tip commit.

Checks run:

- `black --check --line-length 110 --target-version py38` on the changed file — clean. isort was unavailable on this host, but the import grouping (stdlib block, then third-party) and the straight-imports-before-from-imports order match both the `profile = "black"` config and the file's existing shape.
- `ast.parse` on the changed file — clean.
- Confirmed the `pytest` import is still needed: line 44 of the same class uses `@pytest.fixture(autouse=True)`. Autouse fixtures *do* apply to `unittest.TestCase` subclasses — only argument injection does not, which is exactly the asymmetry that produced the bug and why the sibling test in this class was unaffected.
- Swept `tests/` for other `TestCase` subclasses requesting a fixture as an argument — none.

Not verified: the test has not been run green, for the reason in the PR body. This branch's CI cannot execute unit tests, and the test needs torch plus a GPU. Merging still strictly improves the status quo, where the test raises `TypeError` on every invocation. The next MI355X session runs it in-container and reports here.

gphuang:
## Note: test now runs green; the one red check is pre-existing and unrelated

`run-unittest-torch` executed and the checkpoint file reports `test_flux_dist_checkpoint.py ..` — both tests pass. This is the first time the round-trip test has ever run, so #1116's heterogeneous-AdaLN save/load fix now has real regression coverage on a GPU runner rather than a test that raised `TypeError` at call time. That closes the open item in the test plan above.

The job's overall result is still red, from one unrelated failure: `test_primus_turbo_mxfp6_local.py::TestMXFP6LinearModules::test_column_parallel_rejects_tp_gt_1`, which fails at import, not at its assertion.

```
ImportError: cannot import name 'MXFP6_PROLOGUE_BIAS_GELU' from 'primus_turbo.pytorch.core.low_precision'
```

That is a Primus-Turbo version skew in the CI image, not a code problem on this branch: the installed `primus_turbo` predates the symbol `primus_turbo_mxfp6_local.py` imports at line 45. This branch's only change is one test file under `tests/unit_tests/backends/megatron/diffusion/`, so it cannot cause an import failure in an unrelated extension module. The base branch's last run was green on 2026-09-08, so the skew arrived in the image during the intervening week — worth a separate issue, and it will also break any dev container rebuilt from the current image.

Totals: **1 failed, 947 passed, 21 skipped**. Merging, since the alternative is leaving a test that cannot execute in place while an unrelated image skew is sorted out.

PR #1167, opened 2026-09-16
refactor(hybrid): rename Hylo stack back to Zebra-Llama, and fix the hybrid eval/convert tooling

vanshbhatia-amd:
## What

Reverts the `Zebra` -> `Hylo` rename introduced in #1007 (commit `0af1e1a`), restoring **Zebra-Llama** as the name for the hybrid MLA + linear-block stack.

## Why

The 3D world-model upcycling effort is also standardizing on the **Zebra** name. Maintaining two brands for closely related upcycling work is more confusing than helpful, so this consolidates the LLM hybrid stack back under Zebra-Llama.

## Scope

**The name only.** All of the structural improvements from #1007 are kept:

- hybrid configs still carry the mixer and precision in their filenames
- the pure GDN/KDA configs keep their Megatron-style precision names (`gdn_1B_BF16-pretrain.yaml` etc.), untouched here

| Before | After |
|---|---|
| `hylo_llama_<mixer>_<size>_BF16-pretrain.yaml` | `zebra_llama_<mixer>_<size>_BF16-pretrain.yaml` |
| `hylo_<mixer>_<size>_hybrid.yaml` | `zebra_<mixer>_<size>_hybrid.yaml` |
| `hylo_mamba_<size>_hybrid_sft_posttrain.yaml` | `zebra_mamba_<size>_hybrid_sft_posttrain.yaml` |
| `recipes/hylo/hylo.py` | `recipes/zebra/zebra.py` |
| `hylo_llama_flops_patches.py` | `zebra_llama_flops_patches.py` |
| `tools/hybrid/*_hylo_llama*` | `tools/hybrid/*_zebra_llama*` |
| `HyloLlama*` (classes) | `ZebraLlama*` |
| `Primus_Hylo_*` (wandb) | `Primus_Zebra_*` |

## Commits

Two commits, best reviewed separately:

1. `refactor(hybrid): rename Hylo stack back to Zebra-Llama` — the rename, and nothing else. Applying the mechanical `Hylo`/`hylo` to `Zebra`/`zebra` substitution to `main`'s version of each changed file reproduces this commit byte for byte, 71/71.
2. `fix(hybrid): repair the Zebra eval and conversion tooling` — pre-existing defects in the hybrid eval and convert path, raised in review of commit 1. These are equally broken on `main` under the old names; 8 files.

## Implementation

Produced by reverting `0af1e1a` rather than by hand, so the mapping is exactly the inverse of the original rename. The single conflict was `examples/run_pretrain.sh`, which `main` has deleted since; the deletion is preserved.

Two additions on top of the mechanical revert: the MI355X pure GDN/KDA configs landed after #1007 (in #1026 / #1027) and were born with Hylo strings in `wandb_project` and one comment. Those are renamed here as well, so **no Hylo references remain** anywhere in the tree.

## Validation

- `392` insertions / `392` deletions across 76 files, every identifier and file rename paired: no behavioural change
- `git grep -i hylo` returns nothing, and no filename contains `hylo`
- every `model:` reference in every experiment config resolves to an existing model definition
- all 41 changed YAML files parse
- every preset and config path hardcoded in `test_hybrid_model_presets.py` and `test_megatron_trainer.py` exists
- the CI deselect selector matches the renamed test (`test_zebra_mamba_1B_hybrid`)
- all changed Python compiles; no trailing-whitespace or missing-newline lint issues

## Follow-on fixes (commit 2)

Review of commit 1 surfaced genuine defects in the surrounding hybrid tooling. They are not caused by the rename, but since the rename touches these same files they are fixed here rather than left in place:

- **Duplicate `Auto*` registration.** `modeling_zebra_llama.py` registers at import time, so the second registration in `lm_harness_eval.py` and `eval_zebra_llama_lm_eval.sh` raised Transformers' duplicate-registration `ValueError` and aborted both eval entry points before they started.
- **`run_zebra_eval.sh` could not run.** It passed lm-eval's `--model` / `--model_args`, which `lm_harness_eval.py` does not define, so it exited in argparse. Now uses `--model_path` / `--tokenizer` / `--dtype`.
- **The converter now ships its modeling code.** It copies `modeling_zebra_llama.py` into the output directory and records it under `auto_map` in `config.json`, making the documented `trust_remote_code=True` load actually work.
- **Checkpoint paths reconciled.** The converter wrote `iter_0020000` while the chat and eval wrappers looked for `iter_0200000` / `iter_0150000`, so convert-then-chat failed on a missing directory.
- **Docs corrected** for the eval interface, the wrapper's `key=value` argument style, the `{mixer}` filename placeholder, the absence of tokenizer files, and a duplicated word in the technical-guide index that predated #1007.

Additionally validated with `black --check`, `isort --check-only`, `autoflake`, and `bash -n` on every changed wrapper.

## Note for reviewers

This renames public config filenames with no back-compat aliases, matching how #1007 itself removed the legacy aliases. Anyone with pinned `EXP=...hylo_...` paths will need to update them. Happy to add transitional symlinks or aliases if that is preferred.


vanshbhatia-amd:
Rebased onto latest `main`; the conflict is resolved and the PR is mergeable again.

**Conflict.** One file, `docs/06-developer-guide/model-support-matrix.md`. The v26.7 docs update (#1171) extended the Megatron-LM row with the DeepSeek-V4 gfx942 / context-parallelism text on the same line this PR renames. Resolved by keeping that new text and applying only the rename to it, so `Hylo LLaMA` becomes `Zebra LLaMA` and nothing from #1171 is lost.

**On the review findings.** This PR is a pure rename, and it is worth stating that as a checkable property rather than an assertion. Taking `main`'s version of each of the 71 changed files and applying the mechanical `Hylo`/`hylo` to `Zebra`/`zebra` substitution reproduces this branch **byte for byte, 71/71**. The diff is also exactly balanced at 392 insertions and 392 deletions, with every file and identifier rename paired.

That means none of the flagged issues are introduced here; all of them exist on `main` today under the old names. Spot-checking the specific ones:

- `lm_harness_eval.py` accepting `--model_path` rather than `--model` / `--model_args` is unchanged; the docs on `main` have the same mismatch with `--model hylo_llama`.
- The duplicate `AutoModel` registration in the eval wrapper, the `chat_*.py` default checkpoint not matching the converter's output iteration, and the converter not copying `modeling_*.py` into the output directory are all pre-existing behaviours, untouched by the rename.
- `Zebra hybrid hybrid` in `docs/04-technical-guides/README.md` is a real typo, but it predates #1007: `git show 0af1e1a~1:docs/04-technical-guides/README.md` shows the same duplicated word. This branch restores that line verbatim.
- The `{arch}` placeholder omitting the mixer is likewise the pre-existing wording on `main`.

I have deliberately not fixed any of them here, to keep the "renames only, provably no content change" property that makes a 76-file diff reviewable. Happy to follow up with a separate PR for the documentation and eval-wrapper issues, which are genuine and worth fixing on their own merits; say the word if you would rather see them folded into this one.

**Back-compat.** Still no aliases for the old `hylo_*` config paths, matching how #1007 dropped the legacy `zebra_*` aliases. Anyone with pinned `EXP=...hylo_...` paths will need to update. Happy to add transitional aliases if preferred.

vanshbhatia-amd:
Folded the review findings in as a second commit, `68d3a854`. The split is deliberate:

- **`72f00951`** is the rename, and remains a provably pure one. Applying the mechanical `Hylo`/`hylo` to `Zebra`/`zebra` substitution to `main`'s version of each changed file reproduces that commit byte for byte, 71/71, at 392 insertions and 392 deletions.
- **`68d3a854`** is the behavioural fixes, reviewable on its own at 8 files.

Reviewing commit-by-commit is much easier than the squashed diff here.

### What was fixed

**Duplicate `Auto*` registration.** This one was the most serious. `modeling_zebra_llama.py` registers the config and model classes at import time, so the second registration in `lm_harness_eval.py` and in `eval_zebra_llama_lm_eval.sh` raised Transformers' duplicate-registration `ValueError` and aborted both entry points before evaluation began. Both now rely on the import side effect alone. The now-unused `transformers` import was dropped, and the remaining side-effect import is marked `# noqa`, which I confirmed autoflake honours despite `--remove-all-unused-imports`.

**`run_zebra_eval.sh` could not run.** It passed lm-eval's `--model` / `--model_args`, which `lm_harness_eval.py` does not define, so it died in argparse. Now uses `--model_path` / `--tokenizer` / `--dtype`. The tokenizer must be explicit because converted checkpoints contain no tokenizer files.

**The converter now ships the modeling code.** The docs promised the output directory loads with `trust_remote_code=True`, but only weights, config and a model card were written. It now copies `modeling_zebra_llama.py` into the output directory and records it under `auto_map` in `config.json`, so the documented workflow actually works rather than the doc being weakened to match.

**Checkpoint paths disagreed across tools.** The converter wrote `iter_0020000` while `chat_zebra_llama.py` and `run_zebra_eval.sh` looked for `iter_0200000` and `iter_0150000`, so convert-then-chat failed on a missing directory. All four references now agree on the converter's output.

**Docs.** Corrected both eval invocations and the troubleshooting command to the real interface; documented that the shell wrapper takes `key=value` assignments rather than GNU flags and that its task list is fixed in `TASKS`; replaced the inaccurate note claiming the wrapper shells out to `lm_harness_eval.py` with the actual registration path; changed the `{arch}` placeholder to `{mixer}` with its values named, since the real files are `zebra_llama_mamba_*` / `_kda_*` / `_gdn_*`; and removed the duplicated word in the technical-guide index, which predated #1007.

### Deliberately left alone

The `--model_args` usages elsewhere in `tools/hybrid/` and in the GDN and KDA guides are correct: those call the real `lm_eval` CLI or `eval_gdn_lm_eval.py`, which do accept it. Only the `lm_harness_eval.py` call sites were wrong.

Validated with `black --check`, `isort --check-only`, `autoflake`, `bash -n` on every wrapper, `py_compile`, and a parse of all 41 touched YAML files. Still zero `Hylo` references in content or filenames.

Xiaoming-AMD (approved):

PR #1171, opened 2026-09-17
docs(release): update documentation for v26.7

amd-fuyuajin:
## Summary

Documentation for the v26.7 release of both image families. Self-contained — it no
longer stacks on the release-docs tooling PR, so it can merge on its own.

v26.7 moves both families off the ROCm 7.x line onto **ROCm 10.0.0**, which is the
defining change and reaches everything: PyTorch, Transformer Engine, Triton,
torchvision/torchaudio and APEX are all rebuilt against it, the JAX plugin pair is renamed
`jax-rocm10-pjrt` / `jax-rocm10-plugin`, and the wheel indexes move from
`rocm.nightlies.amd.com` to `stable.repo.amd.com`.

- **Release notes** carry both images, with every version read out of the published image
  and cross-checked against the release Dockerfiles. Highlights are split per image
  family, because the release is weighted towards the PyTorch family and one merged list
  implied MaxText gained capability it did not.
- **Version references** bumped from v26.6 to v26.7 across the docs, and checkout
  instructions point at the `release/v26.7` branch now that it exists.
- **Bare-metal install scripts** realigned with the v26.7 Dockerfiles. Four of these
  commits are fixes found by running the installs rather than reading them: wheel indexes
  moved to the stable repo, setuptools is restored after Primus-Turbo downgrades it, the
  Transformer Engine `ck_jit_compile.sh` guard now recognises the upstream fix as well as
  our patch, and both `env.sh` files refuse to install into the other stack's virtualenv.
- **Recipe notes** refreshed per backend, including the MI355X Mamba 370M hipBLASLt
  failure and its `TE_HIPBLASLT_ALGO_SELECTION=1` workaround, both reproduced on hardware.
- **JAX MaxText recipe** drops the retired MAD-integrated benchmarking path, matching what
  #1120 already did for TorchTitan: all training launches through `primus-cli` and nothing
  runs through `examples/*.sh`. The model table was rebuilt from the actual contents of
  `examples/maxtext/configs/`, since the MAD-derived one omitted MI325X, Grok-1,
  Mixtral 8x22B and Llama 3.1 405B.
- **Four link repairs** unrelated to v26.7, broken on main today: `training-recipes.md`
  became `end-to-end-training-recipes.md` and the Megatron-LM patch step was retitled,
  leaving two dead file links and two dead anchors.
- **Management email highlights** under `tools/release_docs/announcements/v26.7.md`,
  deliberately outside `docs/` because ROCm documentation sources `docs/` from this repo
  and that copy is internal.

## Testing

- Every version in the release notes was extracted from the published image rather than
  transcribed, and verified against those image snapshots (124 values across 6/6
  documented images). The verifier itself lives in the release-docs tooling PR, so that
  check cannot be re-run from this branch alone; the tables have not changed since it
  passed.
- In-repo links and anchors: clean, with only the three known unwritten diffusion
  sections outstanding.
- `pre-commit run --all-files` and `tools/ci/check_version_consistency.py` clean.
- Hardware validated: bare-metal installs of both stacks on Ubuntu 22.04 (glibc 2.35) and
  end-to-end training runs on MI355X.

## Follow-ups

- Cherry-pick these docs onto `release/v26.7` after merge, as v26.6 did in #1160.
- `third_party/maxtext` is intentionally **not** bumped here. The image contains
  `b3c53763` and the release notes state that value; the submodule bump itself lands via
  `jax-maxtext-v26.7`.


amd-fuyuajin:
@peterjunpark This is the PR to update the Primus docs to v26.7. Please review it.

GeneDer (approved):
Already tested by @amd-fuyuajin

PR #1172, opened 2026-09-17
Jax docker release - align the JAX backends with v26.7

amd-fuyuajin:
## Summary

Aligns the JAX backends with the v26.7 release. The `tools/perf` batch runner that was
originally on this branch has moved to its own PR so this can land and be cherry-picked
onto `release/v26.7` without waiting on review of a 4k-line feature.

### `third_party/maxtext` → `b3c53763`

Bumps the submodule, points `.gitmodules` at `release/v26.7` instead of `release/v26.6`,
and regenerates `primus/_thirdparty.lock`. This is the MaxText commit the published
`rocm/jax-training:maxtext-v26.7` image actually contains — the v26.7 release notes state
`b3c53763` because the image is the source of truth — so this makes the submodule agree
with what shipped.

Not a fast-forward: `b47d74bf` is not an ancestor of `release/v26.7`, which carries 386
commits the old pin lacks. Two consequences worth a reviewer's eye, both called out in the
commit message: whether the v26.6 backport "fix quantized MoE on the dense_matmul path"
was carried into v26.7, and whether the sliding-window attention fix the gemma4 configs
rely on still behaves the same. The MI325X gemma4 numbers recorded in 85740bb were
measured against v26.6 and have not been re-validated here.

### MaxDiffusion under `primus-cli`

JAX-only TheRock images ship no torch, but MaxDiffusion imports it at module load time
(`max_utils` → `attention_flax` → `modeling_flax_utils` → `modeling_flax_pytorch_utils`)
for the PyTorch-to-Flax weight-conversion helpers, so runs died during config parsing.

The pretrain hook now installs matching ROCm torch and torchvision wheels. The wheels
bundle ROCm runtime libraries, so the wheel source has to match the ROCm in the image, and
the two supported sources are not interchangeable: ROCm >= 10 (TheRock) resolves against
`stable.repo.amd.com` with a `device-<gfx>` extra that reuses the image's installed
`rocm-sdk-device-*` packages, while classic ROCm 7.x uses `repo.radeon.com` find-links.
It also applies the two site-package fixes JAX 0.11 requires: `jnp.clip`'s renamed
`a_min`/`a_max` keywords in Flax T5, and Transformer Engine's empty mesh-axis handling.

Both patches are idempotent, the torch install is skipped when torch is already present,
and the TE patch refuses to apply to an unrecognised Transformer Engine version rather
than corrupting it. Every knob is env-overridable (`MAXDIFFUSION_TORCH_*`).

### Mixtral 8x22B bf16 on MI300X

`per_device_batch_size` 4 → 2. MBS=4 caused OOM

## Testing

- `pre-commit run --all-files` and `tools/ci/check_version_consistency.py` clean.
- The backend config schema drift check runs against the new MaxText pin and reports no
  new drift, so the 386-commit jump introduces no config-schema change.
- Rebased on current `main`.


amd-fuyuajin:
Acting on the split, pushing back on the grouped-bump premise.

**Split: done.** `tools/perf` now lives in its own PR, so this one is just the v26.7
alignment (5 files, +256) and can be cherry-picked onto `release/v26.7` without waiting on
review of the runner.

**`batch-related-dependency-bumps-atomically`: the group does not exist.** There is no
v26.6 release train in `.gitmodules` to bump atomically — before this PR, exactly one
submodule tracked a release branch:

```
$ git show main:.gitmodules | grep 'branch ='
        branch = enable-primus-hybrid-models
        branch = main
        branch = main
        branch = main
        branch = release/v26.6      # third_party/maxtext, the only one
```

The two named as siblings are neither ROCm forks nor on release branches:
`third_party/maxdiffusion` points at `AI-Hypercomputer/maxdiffusion` with no `branch` line
at all, and `third_party/Megatron-LM` points at `NVIDIA/Megatron-LM` tracking `main`. So
moving maxtext to `release/v26.7` leaves nothing behind on a v26.6 pin, and there is no
dependabot/renovate submodule grouping to extend. maxtext is release-branch-tracked
precisely because it is the one submodule whose pin has to match a published image.

On the MaxDiffusion hook "still pinning expectations to `maxtext-v26.6`": that is the only
v26.6 mention, and it is a comment explaining why the patch is idempotent — that image
already carries the guard. It documents observed history rather than pinning anything.

**`no-runtime-package-installation-from-app-code`:** the yq bootstrap moved with
`tools/perf`, so I have answered it in the new PR rather than here.

GeneDer (approved):
Already tested by @amd-fuyuajin

PR #1179, opened 2026-09-18
examples(megatron): bake MXFP4 TE env into YAML env: blocks

coketaste:
## Summary
- `base_env.sh` defaults `NVTE_USE_CAST_TRANSPOSE_TRITON=1`, so CLI-only `primus-cli` launches miss the documented MXFP4 override.
- TrainRuntime already applies top-level `env:` after the GPU files. Put the recipe env there so MAD and `primus-cli` match without extra exports.
- `llama3.1_8B-MXFP4-pretrain.yaml`: `NVTE_USE_CAST_TRANSPOSE_TRITON=0`.
- Flux TE-spec MXFP4: move the old “export these before launch” comment into `env:`.
- YAML `env:` cannot unset `NVTE_FLASH_ATTN` / `NVTE_UNFUSED_ATTN`; the flux comment notes that remaining gap.

## Test plan
- [ ] Confirm TrainRuntime prints `env override: NVTE_USE_CAST_TRANSPOSE_TRITON=0` on llama 8B MXFP4
- [ ] Other MXFP4 recipes (Qwen, flux local-spec) are unchanged; follow-up if they need the same block


coketaste:
Retargeted to `main` (and rebased onto current `main`). `release/v26.7` is a freeze; cherry-pick after merge if this needs to land in that cut.

Xiaoming-AMD (approved):

PR #1180, opened 2026-09-18
fix: honor PRIMUS_SKIP_PIP in the MaxText pretrain install hook

coketaste:
## Summary
- MaxDiffusion already skips pip when `PRIMUS_SKIP_PIP=1` so baked JAX images keep their pinned stack.
- The MaxText pretrain hook on `release/v26.7` still reinstalls `requirements-jax.txt` on every `primus-cli` launch, which can clobber image wheels under MAD.
- Same skip is already on `main`; this backports it to v26.7.

## Test plan
- [ ] `PRIMUS_SKIP_PIP=1 primus-cli direct -- train pretrain --config <maxtext yaml>` logs the skip and does not pip install
- [ ] Default (`PRIMUS_SKIP_PIP` unset) still installs as before


coketaste:
Closing: `PRIMUS_SKIP_PIP` is already on `main` in the MaxText pretrain install hook. This change was only needed as a v26.7 backport, and we are targeting `main` instead of the release freeze.

PR #1181, opened 2026-09-18
fix: apply FLA Triton autotune patch on the primus-cli pretrain path

coketaste:
## Summary
- `examples/run_pretrain.sh` ran `tools/hybrid/patch_fla_triton_autotune_hang.sh` for `hylo_*` / `kda_*` / `gdn_*` configs. `primus-cli` did not, so those models can hang in Triton autotune (`num_stages >= 3`).
- Add `runner/helpers/hooks/train/pretrain/z_patch_fla_triton_autotune.sh` so it runs after `prepare_experiment.sh` (FLA is installed then). Failure is non-fatal, matching the old launcher.

## Test plan
- [x] `bash tests/runner/helpers/test_fla_autotune_hook.sh` (skip path only; matching configs are not executed because the patch mutates installed FLA)
- [x] MAD GDN 1B BF16 on 8× MI350X succeeded with this hook (iter-1 loss/grad matched the old launcher)


coketaste:
Retargeted to `main` (and rebased onto current `main`). `release/v26.7` is a freeze; cherry-pick after merge if this needs to land in that cut.

Xiaoming-AMD (approved):

PR #1178, opened 2026-09-18
fix: add MI350X GPU env file for gfx950 WarpSpeed

coketaste:
## Summary
- `primus-env.sh` sources `runner/helpers/envs/<PRIMUS_GPU_MODEL>.sh` and does **not** fall back to `MI355X.sh`.
- Add `MI350X.sh` with `RCCL_WARP_SPEED_AUTO=0`, matching `MI355X.sh`. Without it, `PRIMUS_GPU_MODEL=MI350X` silently uses only `base_env.sh` and can NaN on gfx950.
- **No change to MI300X/MI325X.** gfx942 atomics stay at the existing `main` / v26.7 defaults (`base_env.sh` + docs/CLI prefixes). Those GPUs are not the v26.7 focus.

## Test plan
- [x] `bash tests/runner/helpers/test_primus_env.sh`

PR #1187, opened 2026-09-24
fix: add MI350X GPU env file for gfx950 WarpSpeed (#1178)

amd-fuyuajin:
## Summary
- `primus-env.sh` sources `runner/helpers/envs/<PRIMUS_GPU_MODEL>.sh` and does **not** fall back to `MI355X.sh`.
- Add `MI350X.sh` with `RCCL_WARP_SPEED_AUTO=0`, matching `MI355X.sh`. Without it, `PRIMUS_GPU_MODEL=MI350X` silently uses only `base_env.sh` and can NaN on gfx950.
- **No change to MI300X/MI325X.** gfx942 atomics stay at the existing `main` / v26.7 defaults (`base_env.sh` + docs/CLI prefixes). Those GPUs are not the v26.7 focus.

## Test plan
- [x] `bash tests/runner/helpers/test_primus_env.sh`

(cherry picked from commit 03e32f72db190303d06fd45eb27a1a5f5e48b17f)


GeneDer (approved):
:shipit:

PR #1188, opened 2026-09-24
examples(megatron): bake MXFP4 TE env into YAML env: blocks (#1179)

amd-fuyuajin:
## Summary
- `base_env.sh` defaults `NVTE_USE_CAST_TRANSPOSE_TRITON=1`, so CLI-only `primus-cli` launches miss the documented MXFP4 override.
- TrainRuntime already applies top-level `env:` after the GPU files. Put the recipe env there so MAD and `primus-cli` match without extra exports.
- `llama3.1_8B-MXFP4-pretrain.yaml`: `NVTE_USE_CAST_TRANSPOSE_TRITON=0`.
- Flux TE-spec MXFP4: move the old “export these before launch” comment into `env:`.
- YAML `env:` cannot unset `NVTE_FLASH_ATTN` / `NVTE_UNFUSED_ATTN`; the flux comment notes that remaining gap.

## Test plan
- [ ] Confirm TrainRuntime prints `env override: NVTE_USE_CAST_TRANSPOSE_TRITON=0` on llama 8B MXFP4
- [ ] Other MXFP4 recipes (Qwen, flux local-spec) are unchanged; follow-up if they need the same block

(cherry picked from commit ad36d5d8e7a1d1dbe61328d0b861bb11214bb60a)


GeneDer (approved):
:shipit:

